PR #4, opened 2022-05-15
add new useClipboard hook, hook up copy address buttons

jakemmarsh:
I couldn't find a corresponding Linear task for this, but let me know if there is one / if one should be made.

- `useClipboard` hook is a simplified, dependency-less version of [this](https://github.com/wsmd/use-clipboard-copy) library
- I opted to use [`navigator.clipboard`](https://caniuse.com/?search=navigator.clipboard), but let me know if browser support is a concern (I would assume IE is a non-issue)
- The interstitial text is just "Copied" right now, happy to change that if there's something else it should be
- I did add one new location that I saw in the Figma mocks but not the current app:

Screen Shot 2022-05-15 at 6 00 57 PM

Let me know if I missed any spots, or if any other new ones should be added!

### DIFF
--- a/packages/client/src/components/SafeContacts.js
+++ b/packages/client/src/components/SafeContacts.js
@@ -1,7 +1,10 @@
 import React from "react";
 import { Person } from "./Svg";
+import useClipboard from "../hooks/useClipboard";
 
 function SafeContacts({ safeOwners = [] }) {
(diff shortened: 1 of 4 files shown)


mannynotfound (approved):
😎

PR #7, opened 2022-05-19
WIP: verify owner addresses and conditionally show check mark

jakemmarsh:


packages/client/src/components/SafeOwners.js
```diff
@@ -63,13 +74,23 @@ function SafeOwners({ address, safeOwners, setSafeOwners }) {
...
-                className="p-4 rounded-sm flex-1"
-                type="text"
-                placeholder="Enter user's FLOW address"
-                value={so?.address}
-                onChange={(e) => onOwnerAddressChange(e.target.value, idx + 1)}
-              />
+              <div className="flex-1" style={{ position: "relative" }}>
+                <input
+                  className="p-4 rounded-sm"
+                  style={{ width: "100%" }}
```
mannynotfound, line 80: could use `column is-full` in className also

packages/client/src/pages/CreateSafe.js
```diff
@@ -101,6 +105,34 @@ function CreateSafe({ web3 }) {
...
+    }
+
+    setSafeOwnersValidByAddress(newSafeOwnersValidByAddress);
+  };
+
+  const onSafeOwnersChange = (newSafeOwners) => {
+    setSafeOwners(newSafeOwners);
+    // skip first safe owner since it's the connected user and won't have an address
+    checkSafeOwnerAddressesValidity(newSafeOwners.slice(1));
+  };
```
mannynotfound, line 135: can we move this into a hook or reducer so it can be shared in other places that need address validation? such as in the transfer tokens flow

packages/client/src/components/SafeOwners.js
jakemmarsh: good to know, updated all instances to do this instead of the hardcoded styles

packages/client/src/pages/CreateSafe.js
jakemmarsh: makes sense, moved into a `useAddressValidation` hook

mannynotfound (approved):
LGTM! Thank you 🙏

PR #12, opened 2022-06-07
VSL-22: hook up address copying and name editing from safe settings

jakemmarsh:
- Added `border-light` to some inputs that weren't matching designs
- Refactored `useClipboard` hook to return a `textJustCopied` value instead of a boolean so that it can be used for multiple values
- Hooked up "Copy" for address in safe settings
- Added new `<EditSafeName />` modal component, hooked up to "Edit" in safe settings

https://user-images.githubusercontent.com/2916671/172436257-5193e30b-f6b7-49c6-a7d3-3b444af3d03c.mov

### DIFF
--- a/packages/client/src/hooks/useTreasury.js
+++ b/packages/client/src/hooks/useTreasury.js
@@ -222,19 +222,19 @@ export default function useTreasury(treasuryAddr, userAddr) {
     return res;
   };
 
-  const setTreasury = (_treasuryAddr, treasuryData) => {
+  const setTreasury = (treasuryAddr, treasuryData) => {
     dispatch({
       type: "SET_TREASURY",
       payload: {
-        [_treasuryAddr]: treasuryData,
+        [treasuryAddr]: treasuryData,
       },
     });
   };
 
-  const fetchTreasury = async (_treasuryAddr) => {
-    const signers = await getSigners(_treasuryAddr);
+  const fetchTreasury = async (treasuryAddr) => {
+    const signers = await getSigners(treasuryAddr);
     if (signers) {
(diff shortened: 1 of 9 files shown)


packages/client/src/hooks/useTreasury.js
```diff
@@ -279,6 +279,16 @@ export default function useTreasury(treasuryAddr, userAddr) {
...
   };
 
+  const renameTreasury = (treasuryAddr, newName) => {
+    dispatch({
+      type: "RENAME_TREASURY",
+      payload: {
+        address: treasuryAddr,
+        name: newName,
+      },
+    });
```
mannynotfound, line 289: I made `SET_TREASURY` a bit generic on purpose so you could be able to just call

```
dispatch({
      type: "SET_TREASURY",
      payload: {
        [address]: {
          name: newName,
        }
    });
```

packages/client/src/pages/Safe.js
```diff
@@ -281,29 +283,31 @@ function Safe({ web3 }) {
...
     );
   };
 
   const onReceive = () => {
     modalContext.openModal(
-      React.createElement(ReceiveTokens, {
-        name: safeData.name,
-        address,
-      })
+      <ReceiveTokens name={safeData.name} address={address} />
```
mannynotfound, line 310: lord knows why I didnt write this in jsx originally but thank you

mannynotfound (commented):
LGTM I think we can just remove redundant reducer action

packages/client/src/hooks/useTreasury.js
jakemmarsh: ah makes sense, removed new action and used this instead

PR #13, opened 2022-06-08
VSL-22: hook up editing of safe owners and threshold for existing treasury

jakemmarsh:
- Moved progress bar logic into a shared `getProgressPercentageForSignersAmount` helper
- Added transactions for updating threshold, updating signer, adding signer
- Added new modals and review flow

https://user-images.githubusercontent.com/2916671/172729909-e9710335-06aa-48de-8643-14ace46cc34b.mov

### DIFF
--- a/packages/client/src/components/SafeSettings.js
+++ b/packages/client/src/components/SafeSettings.js
@@ -445,24 +445,24 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
   const modalContext = useModalContext();
   const safeAddressClipboard = useClipboard();
   const ownersAddressClipboard = useClipboard();
-  const { setTreasury } = web3;
+  const { setTreasury, addSigner, updateThreshold, updateSigner } = web3;
 
   const onEditNameSubmit = (newName) => {
     modalContext.closeModal();
     setTreasury(address, { name: newName });
   };
 
-  const onReviewSafeEditsSubmit = (newOwner, newThreshold) => {
+  const onReviewSafeEditsSubmit = async (newOwner, newThreshold) => {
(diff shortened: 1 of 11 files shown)


packages/client/src/pages/Safe.js
```diff
@@ -84,11 +84,7 @@ const SendTokens = ({ name, address, web3 }) => {
     let newValue = e.target.value;
 
     setRecipient(newValue);
-    let isValid = isAddr(e.target.value);
-    if (isValid) {
-      isValid = await isAddressValid(e.target.value);
-    }
-    setRecipientValid(isValid);
+    setRecipientValid(await isAddressValid(newValue));
```
jakemmarsh, line 87: `isAddressValid` already calls `isAddr` and early-returns if false

packages/client/src/components/SafeSettings.js
```diff
@@ -52,33 +70,442 @@ const EditSafeName = ({ name, onCancel, onSubmit }) => {
...
+    const thresholdToPersist = newThreshold ?? threshold;
+    const ownersToPersist = [...safeOwners];
+
+    if (newOwner) {
+      ownersToPersist.push(newOwner);
+      await addSigner(newOwner.address);
+    }
+
+    if (thresholdToPersist !== threshold) {
+      await updateThreshold(thresholdToPersist);
```
jakemmarsh, line 465: right now we're doing this as a second transaction separate from `addSigner` since you might only change one at a time, but let me know if these should be combined into one transaction that handles the nullability

mannynotfound (approved):
LGTM! Thank you 🙌

PR #18, opened 2022-06-17
VSL-23: treasury transaction history

jakemmarsh:
I'm pretty dubious of the various fields I chose to map to the design and/or the underlying logic for them, so that's probably what mainly needs to be looked at here (or possibly revisited). But this adds the Flowgraph wiring and at least a first pass at transaction history.

| Location      | Screenshot |
| ----------- | ----------- |
| Safe Details      | Screen Shot 2022-06-17 at 4 29 19 PM       |
| Safe Transactions      | Screen Shot 2022-06-17 at 4 29 28 PM       |
| Transaction Details      | Screen Shot 2022-06-17 at 4 29 38 PM       |


mannynotfound (commented):
can we add a "hash" row to link out to flowscan in a new tab 🙏 ?

jakemmarsh:
> can we add a "hash" row to link out to flowscan in a new tab 🙏 ?

added, along with a helper for building the URL according to our current env

Screen Shot 2022-06-23 at 3 10 10 PM

PR #22, opened 2022-06-30
VSL-43: Add Events to VESSEL Contracts

jacksonConrad:
* Adds events to `DAOTreasury.cdc`
* Adds events to `MyMultiSig.cdc`
* Adds events to `TreasuryActions.cdc`

NOTE:

Some of these events are a little redundant, because I wasn't 100% sure which events would be the most useful.

Some of these events also might not contain all the context the front-end needs to query them properly. We will need to play around with the Events API to determine whether we need to enrich any of these events further to serve the FE needs.

### DIFF
--- a/contracts/DAOTreasury.cdc
+++ b/contracts/DAOTreasury.cdc
@@ -7,6 +7,17 @@ pub contract DAOTreasury {
   pub let TreasuryStoragePath: StoragePath
   pub let TreasuryPublicPath: PublicPath
 
+  // Events
+  pub event TreasuryInitialized(initialSigners: [Address], initialThreshold: UInt64)
+  pub event ProposeAction(actionUUID: UInt64)
+  pub event ExecuteAction(actionUUID: UInt64)
+  pub event DepositVault(vaultID: String)
+  pub event DepositCollection(collectionID: String)
+  pub event WithdrawTokens(vaultID: String, amount: UFix64)
+  pub event WithdrawNFT(collectionID: String, nftID: UInt64)
+
+
+  // Interfaces + Resources
   pub resource interface TreasuryPublic {
     pub fun proposeAction(action: {MyMultiSig.Action}): UInt64
     pub fun executeAction(actionUUID: UInt64)
@@ -48,6 +59,7 @@ pub contract DAOTreasury {
     pub fun executeAction(actionUUID: UInt64) {
       let selfRef: &Treasury = &self as &Treasury
       self.multiSignManager.executeAction(actionUUID: actionUUID, {"treasury": selfRef})
+      emit ExecuteAction(actionUUID: actionUUID)
     }
 
     // Reference to Manager //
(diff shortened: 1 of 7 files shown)


generate_accounts.zsh
```diff
@@ -0,0 +1,19 @@
...
+    PUBLICKEY=$(flow keys generate | grep '^Public Key \(.*\)' | tail -c 130 | grep '.*')
+    echo "Private Key:\n$PRIVATEKEY"
+    echo "Public  Key:\n$PUBLICKEY"
+    # Create Account
+    ADDR=$(echo $PUBLICKEY | xargs flow accounts create --key | grep Address | tail -c 19 | grep ".*")
+    echo "Address: $ADDR\n"
+    TIME=1
+    # (echo "emulator-signer$i"; sleep $TIME; echo $ADDR; sleep $TIME; echo; sleep $TIME; echo; sleep $TIME; echo $PRIVATEKEY | xargs; sleep $TIME; echo 0; sleep $TIME) | flow config add account
+    # echo "emulator-signer$i"; sleep $TIME; echo $ADDR; sleep $TIME; echo; sleep $TIME; echo; sleep $TIME; echo $PRIVATEKEY;
+
```
mannynotfound, line 18: can we clean up these comments and make this a normal `.sh` file?

contracts/DAOTreasury.cdc
```diff
@@ -7,6 +7,17 @@ pub contract DAOTreasury {
...
   pub let TreasuryPublicPath: PublicPath
 
+  // Events
+  pub event TreasuryInitialized(initialSigners: [Address], initialThreshold: UInt64)
+  pub event ProposeAction(actionUUID: UInt64)
+  pub event ExecuteAction(actionUUID: UInt64)
+  pub event DepositVault(vaultID: String)
+  pub event DepositCollection(collectionID: String)
+  pub event WithdrawTokens(vaultID: String, amount: UFix64)
+  pub event WithdrawNFT(collectionID: String, nftID: UInt64)
```
mannynotfound, line 17: - can we add an event for a user signing a proposed action?
- can we add an event for an individual nft being sent to treasury?
- can we emit how many tokens were deposited in `DepositVault`?

jacksonConrad: > - can we add an event for a user signing a proposed action?
Yes!

> can we add an event for an individual nft being sent to treasury?
> can we emit how many tokens were deposited in DepositVault?
Currently the way tokens are deposited into the treasury are by calling `borrowVaultPublic()`/`borrowCollectionPublic()`, and then calling the native `deposit` functions on those resource references.  Each token implementation is responsible for calling the `emit Deposit` events, or can choose not to, so it might be hard/inconsistent to capture those events if we rely on those native `Deposit` events.

In order to do this ourselves, we'd need to add specific `depositTokens` and `depositNFT` functions to DAOTreasury.cdc so we can emit our own events.

I'm definitely not against this, but it will need to be added after the Quantstamp audit.
Here is the ticket in Linear: https://linear.app/dappercollectives/issue/VSL-66/implement-depositnft-and-deposittokens-functions-for-treasury

jacksonConrad:
Closing this PR because these changes will be integrated as part of the Quantstamp audit response

PR #24, opened 2022-07-05
VSL-26: Added test case with 20 signers

KsenijaZivkovic:
Based on branch jackson/VSL-43-events

Added util function that creates an array of signers
Added verification for signers in TestTreasurySetup
Added TestTransferTokensToAccountActionsWith20Signers which has the same flow as TestTransferTokensToAccountActions but has 20 signers


KsenijaZivkovic:
Created new branch from the same repo

PR #25, opened 2022-07-05
VSL-26: Added test case with 20 signers

KsenijaZivkovic:
Based on branch jackson/VSL-43-events

Added util function that creates an array of signers
Added verification for signers in TestTreasurySetup
Added TestTransferTokensToAccountActionsWith20Signers which has the same flow as TestTransferTokensToAccountActions but has 20 signers


main_test.go
```diff
@@ -150,6 +155,98 @@ func TestTransferTokensToAccountActions(t *testing.T) {
...
+		assert.Equal(otu.T, uint64(0), treasuryBalance)
+
+		// Assert that all funds have been received by the recipient account
+		recipientBalance := otu.GetAccount(RecipientAcct).Balance
+		assert.Equal(otu.T, TransferAmountUInt64+DefaultAccountBalance, recipientBalance)
+	})
+
+	t.Run("Signers should be able to propose a transfer of a non-fungible token out of the Treasury", func(t *testing.T) {
+		// TODO: create collection in one of the signer accounts
+		otu.ProposeNonFungibleTokenTransferAction("treasuryOwner", MaxSigners[0], "account", uint64(0))
```
mannynotfound, line 213: am I reading this right that the recipient is a string of just "account"? Feels like that should fail lol

jacksonConrad: the recipient in this case is `emulator-account`!  Overflow takes the account name to resolve the address, but you can drop the `emulator-` prefix.

jacksonConrad:
Looks great, approved!  Hold off on merging for now though, please.

jacksonConrad:
closing -- applied changes to `quantstamp-audit/VSL-56_QSP-3`

PR #30, opened 2022-07-06
VSL-50: Added tests for add and remove signers, updated contracts

KsenijaZivkovic:
Based on jackson/VSL-46 branch

Updated transaction to propose actions from the treasury - for add and remove signers
Added happy path tests for both transactions

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -271,3 +271,86 @@ func TestTransferNonFungibleTokensToTreasuryActions(t *testing.T) {
 		assert.Contains(otu.T, ownedNFTIds, uint64(0))
 	})
 }
+
+func TestAddSignerAction(t *testing.T) {
+	var addSignerActionUUID uint64
+
+	otu := NewOverflowTest(t)
+	otu.SetupTreasury("signer1", Signers, uint64(DefaultThreshold))
+
+	t.Run("User should be able to propose a signer to be added", func(t *testing.T) {
+		otu.ProposeAddSignerAction("signer6", "signer1")
+	})
+
+	t.Run("Signers should be able to sign to approve a proposed action to add a new signer", func(t *testing.T) {
+		// Get first ID of proposed action
+		actions := otu.GetProposedActions("signer1")
+		keys := make([]uint64, 0, len(actions))
+		for k := range actions {
+			keys = append(keys, k)
+		}
+		addSignerActionUUID = keys[0]
+
+		// Each signer submits an approval signature
+		for _, signer := range Signers {
+			otu.SignerApproveAction("signer1", addSignerActionUUID, signer)
+		}
+
+		// Assert that the signatures were registered
+		signersMap := otu.GetVerifiedSignersForAction("signer1", addSignerActionUUID)
+		for _, signer := range Signers {
+			assert.True(otu.T, true, signersMap[otu.GetAccountAddress(signer)])
+		}
+	})
+
+	t.Run(`A signer should be able to execute a proposed action to add a signer to Treasury once it has received the required threshold of signatures`, func(t *testing.T) {
+
+		otu.ExecuteAction("signer1", addSignerActionUUID)
+
+		signers := otu.GetTreasurySigners("signer1").String()
+
+		assert.Contains(otu.T, signers, otu.GetAccountAddress("signer6"))
+	})
+
+}
+
+func TestRemoveSignerAction(t *testing.T) {
+	var removeSignerActionUUID uint64
+
+	otu := NewOverflowTest(t)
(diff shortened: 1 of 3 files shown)


main_test.go
```diff
@@ -271,3 +271,86 @@ func TestTransferNonFungibleTokensToTreasuryActions(t *testing.T) {
 		assert.Contains(otu.T, ownedNFTIds, uint64(0))
 	})
 }
+
+func TestAddSignerAction(t *testing.T) {
+	var addSignerActionUUID uint64
+
+	otu := NewOverflowTest(t)
+	otu.SetupTreasury("signer1", Signers, uint64(DefaultThreshold))
```
linxiao-li, line 279: Small nit, probably would be better to keep the convention `treasuryOwner` instead of `singer1`, it helps line 281 read a little better as well ("treasuryOwner proposed adding signer6" instead of "singer1 proposed adding signer6")

main_test.go
```diff
@@ -271,3 +271,86 @@ func TestTransferNonFungibleTokensToTreasuryActions(t *testing.T) {
...
+		assert.Contains(otu.T, signers, otu.GetAccountAddress("signer6"))
+	})
+
+}
+
+func TestRemoveSignerAction(t *testing.T) {
+	var removeSignerActionUUID uint64
+
+	otu := NewOverflowTest(t)
+	otu.SetupTreasury("signer1", Signers, uint64(DefaultThreshold))
```
linxiao-li, line 321: same as above

test_utils.go
```diff
@@ -175,6 +175,26 @@ func (otu *OverflowTestUtils) GetTreasurySigners(account string) cadence.Value {
 	return signers
 }
 
+func (otu *OverflowTestUtils) ProposeAddSignerAction(address, proposingAcct string) *OverflowTestUtils {
+	otu.O.TransactionFromFile("add_signer").
+		SignProposeAndPayAs(proposingAcct).
+		Args(otu.O.Arguments().Address(address)).
```
linxiao-li, line 181: nit, Address as an argument should be indented

test_utils.go
```diff
@@ -175,6 +175,26 @@ func (otu *OverflowTestUtils) GetTreasurySigners(account string) cadence.Value {
...
+		Test(otu.T).
+		AssertSuccess()
+
+	return otu
+}
+
+func (otu *OverflowTestUtils) ProposeRemoveSignerAction(address, proposingAcct string) *OverflowTestUtils {
+	otu.O.TransactionFromFile("remove_signer").
+		SignProposeAndPayAs(proposingAcct).
+		Args(otu.O.Arguments().Address(address)).
```
linxiao-li, line 191: same as above

transactions/add_signer.cdc
```diff
@@ -1,15 +1,16 @@
...
   prepare(signer: AuthAccount) {
-    let treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
                     ?? panic("Could not borrow the DAOTreasury")
-    let manager = treasury.borrowManager()
-    manager.addSigner(signer: additionalSigner)
   }
   execute {
-    return
+    let action =  TreasuryActions.AddSigner(additionalSigner)
```
linxiao-li, line 13: I'm confused here, on the source code TreasuryActions.AddSigner, looks like calling the `AddSinger` function will actually execute and emit the `AddSignerActionExecuted` event. Why proposing the action after the action is executed? tagging @jacksonConrad here as well, please let me know if I misunderstood the flow.

KsenijaZivkovic: I followed the flow for TransferTokenToTreasury.
ProposeAction creates the multisign requirement, so the action cannot be executed without all the necessary signatures
If the action isn't proposed, signed and executed, the treasury won't be updated with the new signer so the assertion will fail

KsenijaZivkovic: So this part is actually creating a struct, and it's not executing an action
`let action =  TreasuryActions.AddSigner(additionalSigner)`
Because AddSigner is defined like this:
`pub struct AddSigner: MyMultiSig.Action`
And after when the action is proposed, signed and executed, the execute() function is called upon this defined action

linxiao-li: Got it! Thanks for the clarification, now it totally makes sense

KsenijaZivkovic:
Closing the PR, because the changes are a part of: https://github.com/DapperCollectives/VESSEL/pull/33/files

PR #31, opened 2022-07-07
VSL-62: Quantstamp Adherence to Best Practices

jacksonConrad:
- Removes duplicate `intent` param from MultiSignAction resource.  Leaves it on MyMultiSig.Action struct
- Remove ZayVerifierv2 contract, as its not needed
- Consolidate `MultiSignAction.verifySignature` function code so that it only iterates over keyIds once
- Remove unnecessary comment in `MyMultiSig.cdc` to improve code readability
- Adds events to all contract state-changing functionality to make it easier for treasury clients & end-users to track treasury state

TODO: Address quantstamp feedback:
```
Currently both verifySignature functions only support ECDSA_secp256k1 and ECDSA_P256 signatures whereas Cadence supports ECDSA_secp256k1, ECDSA_P256, and BLS_BLS12_381 signatures. Add support for BLS_BLS12_381 to increase usability.
```

Its unclear to me whether a BLS signature can be validated the same way that the ECDSA signatures are: https://docs.onflow.org/cadence/language/crypto/#signing-algorithms

We should find a working implementation of BLS signature verification to determine whether it will be compatible with our verifySignature function.


contracts/MyMultiSig.cdc
```diff
@@ -28,7 +38,6 @@ pub contract MyMultiSig {
 
         pub var totalVerified: UInt64
         access(account) var accountsVerified: {Address: Bool}
-        pub let intent: String
         access(contract) let action: {Action}
 
         // ZayVerifierv2 - verifySignature
```
KsenijaZivkovic, line 43: nit: remove comment related to the deleted file

contracts/MyMultiSig.cdc
```diff
@@ -57,74 +66,60 @@ pub contract MyMultiSig {
...
             let account = getAccount(acctAddress)
 
-            let publicKeys: [[UInt8]] = []
-            let weights: [UFix64] = []
-            let signAlgos: [UInt] = []
+            
             
             let uniqueKeys: {Int: Bool} = {}
             for id in keyIds {
                 uniqueKeys[id] = true
```
KsenijaZivkovic, line 73: Maybe we can move the assertion that checks the lengths here to a fail fast approach, for example
```
for id in keyIds {
      assert(uniqueKeys[id] != nil, message: "Invalid duplicates of the same keyID provided for signature")
      uniqueKeys[id] = true
}
```

contracts/MyMultiSig.cdc
```diff
@@ -57,74 +66,60 @@ pub contract MyMultiSig {
...
 
-            var counter = 0
+            // In verify we need a [KeyListSignature] so we do that here
+            let signatureSet: [Crypto.KeyListSignature] = []
             var totalWeight = 0.0
-            while (counter < keyIds.length) {
-                let accountKey: AccountKey = account.keys.get(keyIndex: keyIds[counter]) ?? panic("Provided key signature does not exist")
+
+            var i = 0
+            while (i < keyIds.length) {
```
KsenijaZivkovic, line 82: nit: it's easier to iterate like this, so we don't have to use the keyIds[i]
`for key in keyIds`

contracts/MyMultiSig.cdc
KsenijaZivkovic: oops, now i see that we use signatures[i] too, so might need to ignore this one 😄

contracts/MyMultiSig.cdc
```diff
@@ -57,74 +66,60 @@ pub contract MyMultiSig {
...
+                    weight: keyWeight
                 )
-                i = i + 1
-            }
 
-            // In verify we need a [KeyListSignature] so we do that here
-            let signatureSet: [Crypto.KeyListSignature] = []
-            var j = 0
-            for signature in signatures {
+                var signature = signatures[i]
```
KsenijaZivkovic, line 97: We might need a precondition to check if number of signatures matches the number of keys

contracts/MyMultiSig.cdc
```diff
@@ -57,74 +66,60 @@ pub contract MyMultiSig {
...
                 )
-                j = j + 1
+
+                i = i + 1
             }
 
-            counter = 0
+            assert(totalWeight >= 999.0, message: "Total weight of combined signatures did not satisfy 999 requirement.")
+
             let signingBlock = getBlock(at: signatureBlock)!
```
KsenijaZivkovic, line 110: If there is an error here for getting a block, we could display an intuitive message

flow.json
```diff
@@ -33,7 +33,6 @@
 			}
 		},
 		"ExampleNFT": "./contracts/core/ExampleNFT.cdc",
-		"ZayVerifierV2": "./contracts/core/ZayVerifierv2.cdc",
```
KsenijaZivkovic, line 36: Can we add here the mock signers so we can add the test cases for 20/20 signers?

contracts/MyMultiSig.cdc
KsenijaZivkovic: ignore this one, I see that we are using this variable a lot 😅

PR #33, opened 2022-07-07
VSL-59: QSP-6: Added checks that threshold is never higher than number of signers

KsenijaZivkovic:
Added post and pre checks in MyMultiSig for: 
* removeSigner
* updateThreshold
* init
Added transactions, script and tests for checking if this works as it should

### DIFF
--- a/test_utils.go
+++ b/test_utils.go
@@ -25,7 +25,26 @@ func NewOverflowTest(t *testing.T) *OverflowTestUtils {
 	// return &OverflowTestUtils{T: t, O: overflow.NewOverflowEmulator().Start()}
 }
 
-func (otu *OverflowTestUtils) SetupTreasury(name string, signers []string) *OverflowTestUtils {
+func (otu *OverflowTestUtils) SetupTreasury(name string, signers []string, threshold uint64) *OverflowTestUtils {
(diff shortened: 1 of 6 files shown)


test_utils.go
```diff
@@ -25,24 +25,64 @@ func NewOverflowTest(t *testing.T) *OverflowTestUtils {
 	// return &OverflowTestUtils{T: t, O: overflow.NewOverflowEmulator().Start()}
 }
 
-func (otu *OverflowTestUtils) SetupTreasury(name string, signers []string) *OverflowTestUtils {
+func (otu *OverflowTestUtils) SetupTreasury(name string, signers []string, threshold uint64) *OverflowTestUtils {
 	addresses := make([]string, len(signers))
-	for _, name := range signers {
```
KsenijaZivkovic, line 30: This was buggy, created an array which is double the size because of appending instead of inserting on the right index, and then when the addresses were created based on that - 0x00000000 address was appearing and creating bugs for other tests - especially for adding/removing signers

jacksonConrad: I was wondering why i was seeing the 0x000000 address!! good catch

jacksonConrad (approved):

PR #34, opened 2022-07-07
VSL-60: QSP-7 Allow Signers to Revoke their Approval

jacksonConrad:
- adds `MultiSignAction.signerRevokeApproval` function for signers to revoke their approval of an action
- refactor `MultiSignAction.validateSignature` function to just generally validate whether a message signature is valid.  Validation of the message structure is now the responsibility of the functions that need to validate signatures (i.e. `signerApproveAction` and `signerRevokeApproval`)
- Updates tests + `signer_approve.cdc` tx to adhere to new payload that `signerApproveAction` expects
- Adds `TestSignerRevokeApproval` test

### DIFF
--- a/transactions/signer_approve.cdc
+++ b/transactions/signer_approve.cdc
@@ -2,10 +2,13 @@ import DAOTreasury from "../contracts/DAOTreasury.cdc"
 import MyMultiSig from "../contracts/MyMultiSig.cdc"
 
 transaction(treasuryAddr: Address, actionUUID: UInt64, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
+
+  var isValid: Bool
   let action: &MyMultiSig.MultiSignAction 
   let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
 
   prepare(signer: AuthAccount) {
+    self.isValid = false
     let treasury = getAccount(treasuryAddr).getCapability(DAOTreasury.TreasuryPublicPath)
                     .borrow<&DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}>()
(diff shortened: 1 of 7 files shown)


transactions/signer_revoke.cdc
```diff
@@ -0,0 +1,31 @@
...
+    let treasury = getAccount(treasuryAddr).getCapability(DAOTreasury.TreasuryPublicPath)
+                    .borrow<&DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}>()
+                    ?? panic("A DAOTreasury doesn't exist here.")
+
+    let manager = treasury.borrowManagerPublic()
+    let action = manager.borrowAction(actionUUID: actionUUID)
+
+    var _keyIds: [Int] = []
+
+    var j = 0
```
KsenijaZivkovic, line 16: Is this used?

transactions/signer_approve.cdc
```diff
@@ -19,9 +21,12 @@ transaction(treasuryAddr: Address, actionUUID: UInt64, message: String, keyIds:
...
 
-    var isValid = action.verifySignature(acctAddress: signer.address, message: message, keyIds: _keyIds, signatures: signatures, signatureBlock: signatureBlock)
+    self.messageSignaturePayload = MyMultiSig.MessageSignaturePayload(
+      _signingAddr: signer.address, _message: message, _keyIds: _keyIds, _signatures: signatures, _signatureBlock: signatureBlock
+    )
+
   }
   execute {
-    
+    var isValid = self.action.signerApproveAction(_messageSignaturePayload: self.messageSignaturePayload)
```
KsenijaZivkovic, line 30: Is there a way to return the result of the transaction?
Or the better approach is to add a post block to check if the result is expected

PR #35, opened 2022-07-07
QSP-8: VSL-61: Added input validations for amount and address

KsenijaZivkovic:
Added preconditions for amounts and addresses.
- Amount should be higher than 0.0 in order to be transferred
- Address cannot be address(0)

### DIFF
--- a/contracts/MyMultiSig.cdc
+++ b/contracts/MyMultiSig.cdc
@@ -266,9 +266,8 @@ pub contract MyMultiSig {
         }
 
         pub fun addSigner(signer: Address) {
-            pre {
-                signer != 0x0 : "The signer address should be valid"
-            }
+            // Checks if the signer exists, get account key will fail otherwise
+            getAccount(signer).keys.get(keyIndex: 0) 
             
             self.signers.insert(key: signer, true)
             emit SignerAdded(address: signer)
@@ -356,8 +355,10 @@ pub contract MyMultiSig {
                 _initialSigners.length >= Int(_initialThreshold):
                     "Number of signers must be equal or higher than the threshold."
             }
+
             for signer in _initialSigners {
-                assert(signer != 0x0, message: "The signer address should be valid")
+                // Checks if the signer exists, get account key will fail otherwise
+                getAccount(signer).keys.get(keyIndex: 0)
             }
 
             self.signers = {}
--- a/contracts/MyMultiSig.cdc
+++ b/contracts/MyMultiSig.cdc
@@ -266,6 +266,10 @@ pub contract MyMultiSig {
         }
 
         pub fun addSigner(signer: Address) {
+            pre {
+                signer != 0x0 : "The signer address should be valid"
+            }
+            
             self.signers.insert(key: signer, true)
             emit SignerAdded(address: signer)
         }
@@ -351,6 +355,9 @@ pub contract MyMultiSig {
             pre {
                 _initialSigners.length >= Int(_initialThreshold):
                     "Number of signers must be equal or higher than the threshold."
+            }
+            for signer in _initialSigners {
+                assert(signer != 0x0, message: "The signer address should be valid")
             }
 
             self.signers = {}
--- a/contracts/TreasuryActions.cdc
+++ b/contracts/TreasuryActions.cdc
@@ -60,9 +60,6 @@ pub contract TreasuryActions {
     pub let amount: UFix64
 
     pub fun execute(_ params: {String: AnyStruct}) {
-      pre {
-        self.amount > 0.0 : "Amount should be higher than 0.0"  
-      }
 
       let treasuryRef: &DAOTreasury.Treasury = params["treasury"]! as! &DAOTreasury.Treasury
       let vaultRef: &FungibleToken.Vault = treasuryRef.borrowVault(identifier: self.recipientVault.borrow()!.getType().identifier)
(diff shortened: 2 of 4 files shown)


contracts/MyMultiSig.cdc
```diff
@@ -266,6 +266,10 @@ pub contract MyMultiSig {
         }
 
         pub fun addSigner(signer: Address) {
+            pre {
+                signer != 0x0 : "The signer address should be valid"
```
jacksonConrad, line 270: can you also add the following check:
```
getAccount(signer) != nil
```

contracts/MyMultiSig.cdc
```diff
@@ -352,6 +356,9 @@ pub contract MyMultiSig {
                 _initialSigners.length >= Int(_initialThreshold):
                     "Number of signers must be equal or higher than the threshold."
             }
+            for signer in _initialSigners {
+                assert(signer != 0x0, message: "The signer address should be valid")
```
jacksonConrad, line 360: same here, check `getAccount(signer) != nil`

contracts/TreasuryActions.cdc
```diff
@@ -60,6 +60,10 @@ pub contract TreasuryActions {
     pub let amount: UFix64
 
     pub fun execute(_ params: {String: AnyStruct}) {
+      pre {
+        self.amount > 0.0 : "Amount should be higher than 0.0"
```
jacksonConrad, line 64: i think it might be more appropriate to put this `pre` condition inside the `init` function for these actions.

contracts/TreasuryActions.cdc
```diff
@@ -99,6 +103,10 @@ pub contract TreasuryActions {
...
```
jacksonConrad, line 107: same here, move to `init`

main_test.go
```diff
@@ -152,6 +152,31 @@ func TestTransferTokensToAccountActions(t *testing.T) {
 
 		// TODO: Assert that the NFT has been received by the recipient account
 	})
+
+	t.Run("Signers shouldn't be able to transfer 0.0 fungible tokens out of the Treasury", func(t *testing.T) {
+		otu.ProposeFungibleTokenTransferAction("treasuryOwner", Signers[0], RecipientAcct, 0.0)
```
jacksonConrad, line 157: will need to update the tests here to make sure you can't propose a transfer of 0.0 tokens

main_test.go
```diff
@@ -210,6 +235,30 @@ func TestTransferFungibleTokensToTreasuryActions(t *testing.T) {
 		recipientTreasuryBalance := otu.GetTreasuryVaultBalance("recipient", FlowTokenVaultID)
 		assert.Equal(otu.T, TransferAmountUInt64, recipientTreasuryBalance)
 	})
+
+	t.Run("Signers shouldn't be able to transfer 0 fungible tokens out of the Treasury to another Treasury", func(t *testing.T) {
+		otu.ProposeFungibleTokenTransferToTreasuryAction("treasuryOwner", Signers[0], RecipientAcct, FlowTokenVaultID, 0.0)
```
jacksonConrad, line 240: same here

jacksonConrad (changes requested):

contracts/MyMultiSig.cdc
KsenijaZivkovic: As per discussion, updated this to 
`getAccount(signer).keys.get(keyIndex: 0)`
because it will throw an error if the account doesn't exist

contracts/MyMultiSig.cdc
```diff
@@ -264,8 +264,11 @@ pub contract MyMultiSig {
             emit ActionCreated(uuid: uuid, intent: action.intent)
             return uuid
         }
-
+            
         access(account) fun addSigner(signer: Address) {
+            // Checks if the signer exists, get account key will fail otherwise
+            getAccount(signer).keys.get(keyIndex: 0)
```
jacksonConrad, line 270: it looks like `keys.get` returns an optional value, so i think we need to have
```
assert(getAccount(signer).keys.get(keyIndex: 0) != nil: "Account does not exist").
```

jacksonConrad: or better yet maybe a `pre` condition?

contracts/MyMultiSig.cdc
```diff
@@ -353,6 +356,11 @@ pub contract MyMultiSig {
                     "Number of signers must be equal or higher than the threshold."
             }
 
+            for signer in _initialSigners {
+                // Checks if the signer exists, get account key will fail otherwise
+                getAccount(signer).keys.get(keyIndex: 0)
```
jacksonConrad, line 361: same here, make it an `assert`.

no `pre` since looping through signers in a `pre` condition would be ugly/maybe not possible

jacksonConrad (changes requested):

contracts/MyMultiSig.cdc
KsenijaZivkovic: When I test it, it returns an error
`[Error Code: 1101] cadence runtime error Execution failed:
error: get account key failed: getting account key failed: [Error Code: 1201] account not found for address 0000000000000000
getAccount(signer).keys.get(keyIndex: 0)`

jacksonConrad (approved):

PR #36, opened 2022-07-08
VSL-49: Ensure Treasury can only be destroyed if both vaults and coll…

amyliu6:
…ections are empty

- Added check for vaults and collections before destroy Treasury, as Treasury can only be destroyed if both vaults and collections are empty.


contracts/DAOTreasury.cdc
```diff
@@ -126,9 +126,41 @@ pub contract DAOTreasury {
     }
 
     destroy() {
-      destroy self.multiSignManager
-      destroy self.vaults
-      destroy self.collections
+      // Check if Valuts are empty
+      var isValutEmpty = true
```
jacksonConrad, line 130: slight misspelling, should be `isVaultEmpty`

contracts/DAOTreasury.cdc
```diff
@@ -126,9 +126,41 @@ pub contract DAOTreasury {
...
+      var vaultIds = self.vaults.keys
+      var i = 0
+      while (i < vaultIds.length) {
```
jacksonConrad, line 133: can clean this up a little bit by doing:
```
for identifier in self.vaults.keys {
    let vaultRef = (&self.vaults[identifier] as &FungibleToken.Vault?)!
    if vaultRef.balance > 0.0 {
          isVaultEmpty = false
    }
}
```

contracts/DAOTreasury.cdc
```diff
@@ -126,9 +126,41 @@ pub contract DAOTreasury {
...
+          isValutEmpty = false
+        }
+        i = i + 1
+      }
+
+      // Check if Collections are empty
+      var isCollectionEmpty = true
+      var collectionIds = self.collections.keys
+      var j = 0
+      while (j < collectionIds.length) {
```
jacksonConrad, line 146: same here

transactions/destroy_treasury.cdc
```diff
@@ -0,0 +1,12 @@
+import DAOTreasury from "../contracts/DAOTreasury.cdc"
+
+transaction() {
+  
+  prepare(signer: AuthAccount) {
+    let treasury <- signer.load<@DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    destroy treasury
```
jacksonConrad, line 7: can you move `destroy treasury` to the `execute` block?
will need to define `treasury` variable above prepare block and refer to as `self.treasury`

we are trying to be more intentional with the blocks when writing transactions 😄

contracts/DAOTreasury.cdc
amyliu6: nice catch! 😄

amyliu6:
@jacksonConrad  i just added new commit to address review comments, please help review again. Thanks!

contracts/DAOTreasury.cdc
```diff
@@ -143,9 +143,33 @@ pub contract DAOTreasury {
...
 
     destroy() {
-      destroy self.multiSignManager
-      destroy self.vaults
-      destroy self.collections
+      // Check if Valuts are empty
+      var isVaultEmpty = true
+      for identifier in self.vaults.keys {
+        let vaultRef = (&self.vaults[identifier] as &FungibleToken.Vault?)!
+        if vaultRef.balance > 0.0 {
```
KsenijaZivkovic, line 150: Another approach would be to write it like this:
```
assert(vaultRef.balance > 0.0, message: "Vault " + indentifier + " is not empty!")
```
In this case, the code will fail faster, with a message pointing to the vault that is not empty, instead of going through all the code in destroy

contracts/DAOTreasury.cdc
```diff
@@ -143,9 +143,33 @@ pub contract DAOTreasury {
...
+              isVaultEmpty = false
+        }
+      }
+
+      // Check if Collections are empty
+      var isCollectionEmpty = true
+      for identifier in self.collections.keys {
+        let collectionRef = (&self.collections[identifier] as &NonFungibleToken.Collection?)!
+        if collectionRef.getIDs().length > 0 {
```
KsenijaZivkovic, line 159: Similar would be here
```
assert(collectionRef.getIDs().length > 0, message: "Collection " + indentifier + " is not empty!")
```

amyliu6: 👍

KsenijaZivkovic (approved):

PR #38, opened 2022-07-08
VSL-56: QSP-3 Insufficient Testing

jacksonConrad:


KsenijaZivkovic:
Added commits after PR review and increased coverage for destroyAction, and added edge case for signerRevokeApproval. Also, removed the unused test.cdc

KsenijaZivkovic (approved):

contracts/DAOTreasury.cdc
```diff
@@ -62,6 +64,12 @@ pub contract DAOTreasury {
       emit ExecuteAction(actionUUID: actionUUID)
     }
 
+    pub fun destroyAction(actionUUID: UInt64) {
```
jacksonConrad, line 67: `DestroyAction` should be its own MultiSignAction in `TreasuryActions` and require signers to approve it before executing like other multi-sign actions.  Exposing via `TreasuryPublic` interface would allow anyone to delete a proposed action :)

KsenijaZivkovic: On it!

PR #41, opened 2022-07-14
VSL-70: Add link to remove owner

linxiao-li:
as per the requirments of [this ticket](https://linear.app/dappercollectives/issue/VSL-70/settings-page-add-link-to-remove-owner) this pr:
1. creates the `REMOVE_SIGNER` cadence script to support the action of removing an signer from the safe
2. replaces edit signer UI and logic with remove signer UI and logic
3. adds a `formatAddress` util function to make sure the address passed into cadence starts with `0x`
4. fixed the bug on safe home "things to do" section to display all action items
5. navigate to safe home screen when an action is proposed
6. modified the intent section of addSigner function in multisig to include the address being proposed.


packages/client/src/hooks/useTreasury.js
```diff
@@ -324,6 +335,12 @@ export default function useTreasury(treasuryAddr) {
     await refreshTreasury();
   };
 
+  const removeSigner = async (signerToBeRemovedAddress) => {
+    console.log(signerToBeRemovedAddress)
```
mannynotfound, line 339: we can remove this log statement

packages/client/src/components/SafeSettings.js
```diff
@@ -186,16 +186,15 @@ const EditSignatureThreshold = ({
...
   const [address, setAddress] = useState(safeOwner.address);
   const [addressValid, setAddressValid] = useState(true);
   const isFormValid = useMemo(() => {
     return (
       name.trim().length > 0 &&
-      addressValid &&
-      (name !== safeOwner.name || address !== safeOwner.address)
+      addressValid
     );
   }, [safeOwner, name, address, addressValid]);
```
KsenijaZivkovic, line 199: React warning: React Hook useMemo has unnecessary dependencies: 'address' and 'safeOwner'. Either exclude them or remove the dependency array

amyliu6:
@linxiao-li  could you please update PR title to follow `[ticket number]: [commit message]`? I think that's our current commit message pattern, probably we should work on a contributor guideline for this repo at some point. :)

packages/client/src/components/SafeSettings.js
linxiao-li: fixed, removed the useMemo for reducing unnecessary optimization, here's a blog post explains it well if anyone's intrested: https://kentcdodds.com/blog/usememo-and-usecallback

mannynotfound (approved):
🙏

PR #42, opened 2022-07-14
VSL-66: Implemented depositNFT function for treasury

KsenijaZivkovic:
Removed borrowVault and borrowCollection
Added depositNFT and depositTokens
Updated treasury actions


contracts/DAOTreasury.cdc
```diff
@@ -115,18 +114,30 @@ pub contract DAOTreasury {
       emit DepositCollection(collectionID: identifier)
     }
 
+    // Deposit tokens //
+    pub fun depositTokens(identifier: String, vault: @FungibleToken.Vault) {
+      emit DepositTokens(identifier: identifier)
+
+      let vaultRef = (&self.vaults[identifier] as &FungibleToken.Vault?)!
+      vaultRef.deposit(from: <- vault)
+    }
```
amyliu6, line 123: it looks like we are just changing from ["borrowVault" & "depositVault"] to "depositTokens"? Is it meant to be more readable or what's the purpose of this change?

KsenijaZivkovic: Pinging @jacksonConrad to help here, I'm not sure if this is the right approach
But good catch, depositVault is a really similar to this, so we might just keep one from those two functions

jacksonConrad: good question!  yes, before this PR, the way to deposit tokens into an existing vault was by calling `borrowVaultPublic` and doing `vault.deposit` directly in the transaction.  The downside of this approach is that we cannot emit an `event` for this deposit action when it is done in the transaction.  So the solution is to get rid of `borrowVaultPublic` and replace it with `depositTokens` so we can emit an event, which makes it easier for the Frontend to track.

contracts/DAOTreasury.cdc
```diff
@@ -91,11 +95,6 @@ pub contract DAOTreasury {
       return <- vaultRef.withdraw(amount: amount)
     }
 
-    // Reference to Vault //
-    access(account) fun borrowVault(identifier: String): &FungibleToken.Vault {
-      return (&self.vaults[identifier] as &FungibleToken.Vault?)!
-    }
-
     // Public Reference to Vault //
     pub fun borrowVaultPublic(identifier: String): &{FungibleToken.Balance, FungibleToken.Receiver} {
```
jacksonConrad, line 99: remove `FungibleToken.Receiver` so deposits must go through `depositTokens`

contracts/TreasuryActions.cdc
```diff
@@ -65,8 +65,8 @@ pub contract TreasuryActions {
 
     access(account) fun execute(_ params: {String: AnyStruct}) {
       let treasuryRef: &DAOTreasury.Treasury = params["treasury"]! as! &DAOTreasury.Treasury
-      let vaultRef: &FungibleToken.Vault = treasuryRef.borrowVault(identifier: self.recipientVault.borrow()!.getType().identifier)
-      let withdrawnTokens <- vaultRef.withdraw(amount: self.amount)
+      let collectionID: String = self.recipientVault.borrow()!.getType().identifier
```
jacksonConrad, line 68: nit: rename `collectionID` -> `vaultID`

jacksonConrad (changes requested):
Just two small changes :)

Also is the front-end work related?  No problem with it, just curious.

KsenijaZivkovic:
> Just two small changes :)
> 
> Also is the front-end work related? No problem with it, just curious.

FE work is not related, but I figured it would be a nice improvement to display thumbnails from the minted NFTs. I can also create a ticket and move those unrelated changes to a new PR, it will be cleaner to review both of them.

contracts/DAOTreasury.cdc
KsenijaZivkovic: Updated, please check if the change is correct

contracts/DAOTreasury.cdc
```diff
@@ -116,10 +103,13 @@ pub contract DAOTreasury {
 
     // Deposit tokens //
     pub fun depositTokens(identifier: String, vault: @FungibleToken.Vault) {
-      emit DepositTokens(identifier: identifier)
+      if self.vaults[identifier] != nil {
+        self.vaults[identifier]?.deposit!(from: <- vault)
+      } else {
+        self.vaults[identifier] <-! vault
```
jacksonConrad, line 109: in the case there isn't already a vault here, we should throw an error.  maybe add an assert to check this at the beginning of this function

jacksonConrad (changes requested):
just caught one thing that i missed last time :)

contracts/DAOTreasury.cdc
KsenijaZivkovic: good catch!

KsenijaZivkovic: But if we throw an error there will be no possibility to add a vault with a different identifier?

linxiao-li (commented):
LGTM, just curious what's the thoughts behind getting rid of the borrow collection and borrow vault in the exploit tests

KsenijaZivkovic:
> LGTM, just curious what's the thoughts behind getting rid of the borrow collection and borrow vault in the exploit tests

borrowVault and borrowCollection are deleted functions, so there is no way to call them in these exploit transactions, so I had to remove them

contracts/DAOTreasury.cdc
KsenijaZivkovic: Fixed!

contracts/DAOTreasury.cdc
KsenijaZivkovic: Per discussion from today: Reverted depositVault function to be accessible to the resource owner, and added to create_treasury.cdc initial setup of the flow vault

jacksonConrad (approved):

PR #43, opened 2022-07-15
VSL-46: Signers may add new collections to treasury

KsenijaZivkovic:
- depositVault is not exposed via TreasuryPublic interface anymore because we are allowing only the owner of the resource to use it
- Added signer payload to depositCollection treasury function
- Updated tests and UI to work with this change
- Updated nft.js - to only do the minting of NFT, since deposit collection button on UI is functional
First step when clicking on depositing a collection button (costs no gas):
Screen Shot 2022-07-21 at 12 57 35 PM
Next step is to authorize this transaction:
Screen Shot 2022-07-21 at 12 58 29 PM


contracts/DAOTreasury.cdc
```diff
@@ -108,6 +147,54 @@ pub contract DAOTreasury {
 
     // ------- Collections ------- 
 
+    pub fun signerDepositCollection(collection: @NonFungibleToken.Collection, signaturePayload: MyMultiSig.MessageSignaturePayload) {
+      // ------- Validate Address is a Signer on the Treasury -----
+      let signers = self.multiSignManager.getSigners()
+      assert(signers[signaturePayload.signingAddr] == true, message: "Address is not a signer on this Treasury")
+
+      // ------- Validate Message --------
+      // message format: {vault identifier hex}{blockId}
```
KsenijaZivkovic, line 156: collection instead of vault

main_test.go
```diff
@@ -54,14 +54,13 @@ func TestTreasurySetup(t *testing.T) {
 	})
 
 	t.Run("Treasury should be able to receive non-fungible tokens", func(t *testing.T) {
-		otu.CreateNFTCollection("account")
-		otu.MintNFT("account")
+		otu.CreateNFTCollection("signer1")
+		otu.MintNFT("signer1")
```
jacksonConrad, line 58: whats the purpose of changing up the account names here?

transactions/send_flow_to_treasury.cdc
```diff
@@ -11,7 +11,8 @@ transaction(treasuryAddr: Address, amount: UFix64) {
                     .borrow<&DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}>()
                     ?? panic("A DAOTreasury doesn't exist here.")
 
-    
-    pubCapability.depositVault(vault: <- tokens)
+    let treasuryVault: &{FungibleToken.Balance, FungibleToken.Receiver} = pubCapability.borrowVaultPublic(identifier: vault.getType().identifier)
```
jacksonConrad, line 14: will need to update this to `depositTokens` when that PR is merged

main_test.go
KsenijaZivkovic: I think I got lost in tests, will revert to account and then see if it still works :D once the previous PR is merged I'll update the changes

jacksonConrad: sounds good! this might all work, just seemed unnecessarily confusing

KsenijaZivkovic: Only signers can deposit NFT collections, so I had to update it after all to signer1

transactions/create_treasury.cdc
```diff
@@ -8,8 +9,10 @@ transaction(initialSigners: [Address], initialThreshold: UInt64) {
 
     // Seed Treasury with commonly used vaults
     let flowVault <- FlowToken.createEmptyVault()
+    let fusdVault <- FUSD.createEmptyVault()
```
amyliu6, line 12: Can we update FE as well to create all supported Vault FLOW/FUSD/USDC ? https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/flow/initializeTreasury.tx.js#L11

KsenijaZivkovic: Can it be done as part of https://linear.app/dappercollectives/issue/VSL-52/add-fusd-usdc-flow?

amyliu6: sure, that sounds good! 👍

jacksonConrad (approved):

amyliu6 (approved):

PR #44, opened 2022-07-15
[DO NOT MERGE] VSL-73: Add Dapper Wallet as Authentication Option

jacksonConrad:
* Opt-in to Dapper Wallet via FCL config

NOTE: Do not merge until
- Dapper Wallet supports signed messages
- We have received clearance from compliance to integrate Dapper Wallet into VESSEL


jacksonConrad:
closing because Dapper Wallet doesnt support message signing yet so will hold off until they do.

PR #47, opened 2022-07-19
VSL-85: Updated metadata and transactions for NFT UI

KsenijaZivkovic:
- Updated nft.js with a thumbnail url
- Updated transactions to return not only ids of nfts in collections, but all other data
- Thanks to this, it is possible to display name, description and thumbnail on the UI
- Example:
Screen Shot 2022-07-19 at 11 02 54 AM
- After depositing the collection you can run **yarn nft** in order to mint an NFT and send it to treasury, so this cute cat can be visible in the Assets tab
- TODO for the future: update Deposit NFT to show a screen in order to select which NFT we want to send to the treasury (not part of this improvement)


amyliu6 (approved):
LGTM!

PR #49, opened 2022-07-20
vsl-70: post add remove signer render fix

linxiao-li:
This is a temporary fix for updating the owners list after an add/remove signer proposal has been approved. Demo for the flow of adding a safe owner:
1. After adding the safe owner on the setting screen, user should be navigated to the home screen
2. An action item should be created for the user to sign and confirm on the "Things to do" section
3. After signing and confirming the proposal, the action item should disappear from the thing to do list
4. Navigate back to the safe settings page, the new owner should be added to the owners list

https://user-images.githubusercontent.com/7423845/180272685-dd16a5fe-0362-46d8-b57a-76da170088e6.mov

Created a tech debt ticket here :https://linear.app/dappercollectives/issue/ENG-745/[tech-debt]-clean-up-safeowners-data-object-and-have-signers-as-single


amyliu6:
@linxiao-li  Thanks for creating ticket to track improvement work! 👍 
Could you help attach a screenshot/gif to show case the UI work under PR description? that would help provide more context on what's been fixed here.

packages/client/src/components/SafeSettings.js
```diff
@@ -192,7 +192,9 @@ const RemoveSafeOwner = ({ web3, safeOwner, onCancel, onSubmit }) => {
   const [name, setName] = useState(safeOwner.name);
   const [address, setAddress] = useState(safeOwner.address);
   const [addressValid, setAddressValid] = useState(true);
-  const isFormValid =  name.trim().length > 0 && addressValid;
+  //LL: will validate name once name is persisted on the chain
```
amyliu6, line 195: so that means a contract change? what's the plan for that work?

linxiao-li: Yes, that means we need to change the data structure of signers from {Address: Bool} (representing if the signer is verified) to {Address: SignerStruct} (SignerStruct contains multiple fields like name, isVerified, etc). I believe this change is necessary for the work moving forward to handle users identities. Regardless if we want to use an external api for user identities or not, signer object needs to be more descriptive. 

The work mainly lies in tweaking the MyMultiSig.cdc, I don't see any script/transactions or front-end being impacted so far.

amyliu6: I see. Noticed that in this [Vessel PRD page](https://www.notion.so/VESSEL-Treasury-Management-49d5dcb8b03d41898d273af0f79dee87) it mentioned ```All naming and imagery added to a safe is saved locally. Nothing is shared amoong wallet owners on safe besides address.```, wondering if it's per design to not have name on chain? 🤔  @jacksonConrad any idea?

KsenijaZivkovic:
LGTM! Can you also format the changes in the files?

KsenijaZivkovic (approved):

packages/client/src/components/SafeSettings.js
amyliu6: Just to close on this thread, there has been offline discussion on this and confirmed it's expected to not have `name` stored on chain. In future we might consider store it in a service but for now it will be on local storage only.

amyliu6 (approved):
LGTM! Just a reminder to update PR title `<ticket number>: <message>` :)

mannynotfound (approved):
🙌

PR #50, opened 2022-07-22
VSL-80: Propose & Execute actions can be only done by signers

KsenijaZivkovic:
- Extended Action to include proposer
- Updated actions to require signature in order to verify that the signer of the treasury can propose and execute actions
- Updated all transactions to require additional parameters in order to verify the signer
- Updated the UI to work with these changes 
- Updated the tests that cover the updated transactions


packages/client/src/components/SafeSettings.js
```diff
@@ -447,13 +447,62 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
...
+      const keyIds = [sigResponse[0]?.keyId];
+      const signatures = [sigMessage];
+
+      await updateThreshold(
+        thresholdToPersist,
+        message,
+        keyIds,
+        signatures,
+        height
+      );
```
linxiao-li, line 505: same as above

packages/client/src/components/SafeSettings.js
```diff
@@ -467,7 +516,31 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
...
+      const keyIds = [sigResponse[0]?.keyId];
+      const signatures = [sigMessage];
+
+      await proposeRemoveSigner(
+        formatAddress(ownerToBeRemoved.address),
+        message,
+        keyIds,
+        signatures,
+        height
+      );
```
linxiao-li, line 543: Since this part of the logic appears 4 times in our code, can we extract it into a utility function?

packages/client/src/hooks/useTreasury.js
```diff
@@ -50,15 +50,27 @@ const doSendFlowToTreasury = async (treasuryAddr, amount) => {
...
       arg(treasuryAddr, t.Address),
       arg(recipientAddr, t.Address),
       arg(amount, t.UFix64),
+      arg(message, t.String),
+      arg(keyIds, t.Array(t.UInt64)),
+      arg(signatures, t.Array(t.String)),
+      arg(height, t.UInt64),
     ],
-    limit: 55,
+    limit: 300,
```
amyliu6, line 73: How do we come up with this value? Can we consider define constant to hold the limit value?

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Yeah, makes sense, a significant part is fixed so it can be moved to utility

packages/client/src/hooks/useTreasury.js
KsenijaZivkovic: Gas limit is used up for all these transactions that require signature validation (because there are a couple of checks needed), so I had to raise this value - first to 250 and then to 300 (because for this one this limit is not enough). There is a [question on slack](https://brud.slack.com/archives/C03ERT4F2MB/p1658363099458749) about what are the recommendations for these limits.
For the other question - yes, this can be defined as a constant - and for all the transactions that have a signature validation this limit can be the same, so I will introduce that here. Thank you for the input!

KsenijaZivkovic: It depends on the computationUsed - which can be read from flow emulator when the transaction is executed, for each transaction it is different, and for this one it's around 280

packages/client/src/components/SafeSettings.js
```diff
@@ -447,13 +447,62 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
...
+      const message = `${newOwnerHex}${id}`;
+      const messageHex = Buffer.from(message).toString("hex");
+
+      let sigResponse = await web3.injectedProvider
+        .currentUser()
+        .signUserMessage(messageHex);
+      const sigMessage =
+        sigResponse[0]?.signature?.signature ?? sigResponse[0]?.signature;
+      const keyIds = [sigResponse[0]?.keyId];
+      const signatures = [sigMessage];
```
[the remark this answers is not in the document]
KsenijaZivkovic: Updated

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Moved to util

packages/client/src/hooks/useTreasury.js
KsenijaZivkovic: Added constants

packages/client/src/contexts/Web3.js
```diff
@@ -4,6 +4,13 @@ import networks from "../networks";
 import { useRouteMatch } from "react-router-dom";
 import { useFclUser, useTreasury, useNFTs } from "../hooks";
 
+export const REGULAR_LIMIT = 55;
+export const CREATE_TREASURY_LIMIT = 80;
+export const UPDATE_SIGNER_LIMIT = 110;
+export const SIGNED_LIMIT = 300;
```
amyliu6, line 10: just curious for those actions with SIGNED_LIMIT, are they sharing common complexity of contract code? If there is any chance that certain action need to adjust limit but not for others, we might want to separate them now so to have the flexibility for limit update in future.

KsenijaZivkovic: Proposing all actions has the same checks in the contract code (is the signature valid, is it the signer on the treasury etc), and similar computation effort is used, so the similar limit should be there

packages/client/src/components/SafeSettings.js
```diff
@@ -447,11 +448,29 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
     const thresholdToPersist = newThreshold ?? threshold;
 
     if (newOwner) {
-      await proposeAddSigner(formatAddress(newOwner.address));
+      const intent = `Add account ${newOwner.address} as a signer.`;
+      const { message, keyIds, signatures, height } = createSignature(web3, intent);
```
amyliu6, line 452: Instead of defining params here and pass through multiple methods to reach the one in `useTreasury.js`, can we consider define the params inside `useTreasury.js` itself?

KsenijaZivkovic: Makes sense, will update it! :)

KsenijaZivkovic: Updated, please take a look again when you have time

contracts/DAOTreasury.cdc
```diff
@@ -50,10 +53,57 @@ pub contract DAOTreasury {
...
+    pub fun executeAction(actionUUID: UInt64, signaturePayload: MyMultiSig.MessageSignaturePayload) {
+      self.validateTreasurySigner(identifier: actionUUID.toString(), signaturePayload: signaturePayload)
+
       let selfRef: &Treasury = &self as &Treasury
       self.multiSignManager.executeAction(actionUUID: actionUUID, {"treasury": selfRef})
-      emit ExecuteAction(actionUUID: actionUUID)
+      emit ExecuteAction(actionUUID: actionUUID, proposer: signaturePayload.signingAddr)
+    }
+
+    access(self) fun validateTreasurySigner(identifier: String, signaturePayload: MyMultiSig.MessageSignaturePayload) {
```
jacksonConrad, line 64: can we remove this fn and just use MyMultiSig.verifySignature?

contracts/DAOTreasury.cdc
```diff
@@ -50,10 +53,57 @@ pub contract DAOTreasury {
...
       to know what they are signing.
     */
-    pub fun executeAction(actionUUID: UInt64) {
+    pub fun executeAction(actionUUID: UInt64, signaturePayload: MyMultiSig.MessageSignaturePayload) {
+      self.validateTreasurySigner(identifier: actionUUID.toString(), signaturePayload: signaturePayload)
+
       let selfRef: &Treasury = &self as &Treasury
       self.multiSignManager.executeAction(actionUUID: actionUUID, {"treasury": selfRef})
-      emit ExecuteAction(actionUUID: actionUUID)
+      emit ExecuteAction(actionUUID: actionUUID, proposer: signaturePayload.signingAddr)
```
amyliu6, line 61: could you help me understand why there is a need to add proposer to Action, if proposer is already available in `signaturePayload.signingAddr`?
https://github.com/DapperCollectives/VESSEL/pull/50/files#diff-d6c67e336e7c8eb3ba7d49cd0a16be12910ac98006b41b6abfddf320a36295f1R31

packages/client/src/hooks/useTreasury.js
```diff
@@ -39,27 +47,43 @@ const doCreateTreasury = async (signerAddresses, threshold) => {
...
 
-const doProposeTransfer = async (treasuryAddr, recipientAddr, amount) => {
+const doProposeTransfer = async (
+  web3,
+  treasuryAddr,
+  recipientAddr,
+  amount
+) => {
+  const uFixAmount = String(parseFloat(amount).toFixed(8));
+  const tokenAddress = process.env.REACT_APP_FLOW_ENV === "emulator" ? "ee82856bf20e2aa6" : "9a0766d93b6608b7";
```
amyliu6, line 69: is this for testing only?

contracts/MyMultiSig.cdc
```diff
@@ -28,6 +28,7 @@ pub contract MyMultiSig {
```
amyliu6, line 28: Noticed this intent params seems not in use, if so can we clean it up in a separate PR? https://github.com/DapperCollectives/VESSEL/blob/05a324ece69f212a0884759352ad573f15bb63c3/contracts/MyMultiSig.cdc#L162

packages/client/src/components/SafeSettings.js
```diff
@@ -447,11 +447,11 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
     const thresholdToPersist = newThreshold ?? threshold;
 
     if (newOwner) {
-      await proposeAddSigner(formatAddress(newOwner.address));
+      await proposeAddSigner(web3, formatAddress(newOwner.address));
```
amyliu6, line 450: can we access `web3` like in @linxiao-li 's PR: https://github.com/DapperCollectives/VESSEL/pull/52/files#diff-55a64a22c46f532689008bceccb29906ccc8380b87ccd064ec508835cad48958R12

linxiao-li:
Since 'useTreasury' hook is a dependency of the Web3Context, bringing web3 as an argument to the api functions defined in useTreasury could be problematic. 
Consider moving the `createSignature` function from the util to the web3 context to utilize the context within, that way we can also avoid passing web3 (which is a big object) through multiple functions.

jacksonConrad:
Contracts look good to me but we should hold off on merging until we can deploy the updated contracts (which we may need to rename)

packages/client/src/hooks/useTreasury.js
KsenijaZivkovic: This action has an intent that contains the identifier of a recipient vault, which contains information about what account is the FungibleToken contract deployed to, so this is a "hacky" solution to sign the message in a correct way if we keep the intent as is.
TransferToken action should work for every type of FungibleToken that is available, so ideally intent should contain something readable, like $FLOW, $USDC, $FUSD... and this is also necessary for the UI when user proposes the action to transfer tokens from the treasury (to avoid hardcoding or changing the values that come from BE)

contracts/MyMultiSig.cdc
KsenijaZivkovic: Sure, let's create a ticket for it

contracts/DAOTreasury.cdc
KsenijaZivkovic: Might be useful in case we want in the future to display who proposed the action. For example script get_proposed_actions.cdc returns the intents, this might be expanded to support the proposers too. ExecuteAction here is an event (naming can be confusing I know :D) that is emitted only once when the action is successfully executed
Looping in @jacksonConrad for thoughts

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Moved createSignature function from util to web3 context

contracts/DAOTreasury.cdc
amyliu6: I see. As the ticket is related to restrict action to signers, I was wondering why we need to add `proposer` to Action. So looks like its two different items - Adding proposer to action is purely extending info in Action. :)

packages/client/src/hooks/useTreasury.js
amyliu6: got it. So we'll need to pass the token type based on user's selection and maintain an address mapping to be able to dynamically set here. Do we already have a ticket to cover this work? If no I can create one. cc @linxiao-li

contracts/MyMultiSig.cdc
amyliu6: Could you please share the ticket link here? Let me know if you need help to create one.

amyliu6 (approved):
@KsenijaZivkovic  Thanks for being patient with multiple reviews and iterating on this PR! 👍

packages/client/src/hooks/useTreasury.js
linxiao-li: We can put the mapping information in `src/constants/maps.js`. I'll work with @KsenijaZivkovic to compose the map

linxiao-li (approved):

linxiao-li:
Looks like there's a conflict caused by the modal refactor pr, sendTokens.js is safe to remove 👍

contracts/MyMultiSig.cdc
KsenijaZivkovic: https://linear.app/dappercollectives/issue/VSL-109/multisignaction-cleanup

packages/client/src/flow/addSigner.tx.js
```diff
@@ -1,18 +1,31 @@
...
     
-    let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
-
+    let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+    let action: AnyStruct{MyMultiSig.Action}
+    let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
+  
     prepare(signer: AuthAccount) {
-      self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+      self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 13: We should add the treasury Address to the args of this transaction.  This `signer.borrow` will only work if the Treasury resource is stored in the signer's storage.

packages/client/src/flow/removeSigner.tx.js
```diff
@@ -1,18 +1,30 @@
...
+transaction(signerToBeRemoved: Address, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
   
-  let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let action: AnyStruct{MyMultiSig.Action}
+  let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
 
   prepare(signer: AuthAccount) {
-    self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 13: same here -- need treasuryAddr

packages/client/src/flow/updateThreshold.tx.js
```diff
@@ -1,18 +1,31 @@
...
-    let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  transaction(newThreshold: UInt64, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
+    
+    let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+    let action: AnyStruct{MyMultiSig.Action}
+    let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
   
     prepare(signer: AuthAccount) {
-      self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+      self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 13: add treasuryAddr

transactions/add_signer.cdc
```diff
@@ -1,16 +1,29 @@
...
+transaction(additionalSigner: Address, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
   
-  let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let action: AnyStruct{MyMultiSig.Action}
+  let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
 
   prepare(signer: AuthAccount) {
-    self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 12: add treasuryAddr

transactions/remove_signer.cdc
```diff
@@ -1,16 +1,28 @@
...
+transaction(signerToBeRemoved: Address, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
   
-  let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let action: AnyStruct{MyMultiSig.Action}
+  let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
 
   prepare(signer: AuthAccount) {
-    self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 12: add treasuryAddr

transactions/update_threshold.cdc
```diff
@@ -1,16 +1,29 @@
...
+transaction(newThreshold: UInt64, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
   
-  let Treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let treasury: &DAOTreasury.Treasury{DAOTreasury.TreasuryPublic}
+  let action: AnyStruct{MyMultiSig.Action}
+  let messageSignaturePayload: MyMultiSig.MessageSignaturePayload
 
   prepare(signer: AuthAccount) {
-    self.Treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
+    self.treasury = signer.borrow<&DAOTreasury.Treasury>(from: DAOTreasury.TreasuryStoragePath)
```
jacksonConrad, line 12: add treasuryAddr

jacksonConrad (changes requested):
Requested changes for a somewhat unrelated bug in some of the transactions.

Can update on this ticket, or break out into separate ticket :)

jacksonConrad (approved):

PR #51, opened 2022-07-25
VSL-91: Added a pre check if signer already exists before adding

KsenijaZivkovic:
- This check is triggered when the action is executed
- Added check on UI against adding the same user to the treasury


packages/client/src/components/SafeSettings.js
```diff
@@ -266,9 +265,14 @@ const AddSafeOwner = ({ web3, onCancel, onNext }) => {
 
   const onAddressChange = async (newAddress) => {
     setAddress(newAddress);
-    setAddressValid(isAddr(newAddress) && (await isAddressValid(newAddress)));
+    setAddressValid(isAddr(newAddress) && (await isAddressValid(newAddress)) && !isAddressExisting(safeOwners, newAddress));
   };
 
+  const isAddressExisting = (safeOwners, newAddress) => {
+    const address = newAddress.startsWith("0x") ? newAddress : `0x${newAddress}`;
```
amyliu6, line 272: we can use this `formatAddress` defined here: https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/utils.js#L45

KsenijaZivkovic: Nice catch, updated!

packages/client/src/components/SafeSettings.js
```diff
@@ -193,7 +193,6 @@ const EditSignatureThreshold = ({
 
 const RemoveSafeOwner = ({ web3, safeOwner, onCancel, onSubmit }) => {
   const { isAddressValid } = useAddressValidation(web3.injectedProvider);
-  const [name, setName] = useState(safeOwner.name);
   const [address, setAddress] = useState(safeOwner.address);
```
linxiao-li, line 196: Good job removing the unnecessary name state here!

packages/client/src/components/SafeSettings.js
```diff
@@ -266,9 +265,14 @@ const AddSafeOwner = ({ web3, onCancel, onNext }) => {
 
   const onAddressChange = async (newAddress) => {
     setAddress(newAddress);
-    setAddressValid(isAddr(newAddress) && (await isAddressValid(newAddress)));
+    setAddressValid(isAddr(newAddress) && (await isAddressValid(newAddress)) && !isAddressExisting(safeOwners, newAddress));
   };
 
+  const isAddressExisting = (safeOwners, newAddress) => {
+    const address = formatAddress(newAddress);
+    return safeOwners.filter(obj => obj.address === address).length !== 0;
```
linxiao-li, line 273: We should make sure the `verified` field gets evaluated as well
```suggestion
    return safeOwners.filter(obj => obj.address === address && obj.verified).length !== 0;
```

KsenijaZivkovic: Added!

jacksonConrad (approved):

linxiao-li (approved):

PR #53, opened 2022-07-28
VSL-106: Updated versions so overflow testing works with emulator

KsenijaZivkovic:
Updated go to 1.18.0
Updated overflow, cadence, flow-go-sdk and other dependencies
Updated transactions and tests as a result from updating libraries

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,12 +1,12 @@
 module github.com/DapperCollectives/VESSEL
 
-go 1.17
+go 1.18
 
 require (
-	github.com/bjartek/overflow v0.0.0-20220625182952-d42d54fb53b4
-	github.com/onflow/cadence v0.24.3
-	github.com/onflow/flow-go-sdk v0.26.2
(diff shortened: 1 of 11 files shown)


linxiao-li:
LGTM
Do the updated libraries require arguments being passed explicitly? For example from `RemoveSigner(signerToBeRemoved)` to `RemoveSigner(_signer:signerToBeRemoved)`

amyliu6 (approved):

KsenijaZivkovic:
> LGTM Do the updated libraries require arguments being passed explicitly? For example from `RemoveSigner(signerToBeRemoved)` to `RemoveSigner(_signer:signerToBeRemoved)`

Yeah, the cadence library that was referenced was asking to add these arguments to be passed. Note: I had to downgrade it a bit in order to work with the overflow library, but we might need some updates of dependencies again soon so when we return to the latest version of cadence these arguments can be removed

jacksonConrad (approved):

PR #58, opened 2022-07-30
VSL-52: Added USDC (FiatToken)

KsenijaZivkovic:
## Description
Added FiatToken and OnChainMultiSig to core contracts
Created a deploy.go script that is enabling this contract to be deployed to the emulator and enable local development
Added USDC vault to treasury when the treasury is created
Updated tests to deploy this script before running the tests

## Demo / Test Result
UI looks the same, but I've added one console.log when safe is created and each time it is loaded, to show what types of vaults are created with the create_treasury transaction:
Screen Shot 2022-08-02 at 3 48 37 PM
In the picture, first vault is for Flow tokens, second is for FUSD and third is for FiatToken (USDC)


go/deploy.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+type Addresses struct {
+	FungibleToken   string
+	OnChainMultiSig string
+}
+
+func main() {
+	otu := DeployOverflow()
+	otu.DeployFiatTokenContract("treasuryOwner", "USDC", "0.1.0")
```
amyliu6, line 27: could you help me understand why USDC require special script to deploy? and is this script only does deploy or there are other setups we are trying to achieve?

KsenijaZivkovic: USDC init method has a lot of parameters, and it cannot be deployed over flow.json, so the script needed to be added.
I took this and modified it in order to work https://github.com/flow-usdc/flow-usdc/blob/main/lib/go/deploy/USDC.go
This script does the deployment of all contracts and the USDC contract too, and it's only for local testing, because on testnet this is already deployed (it can be seen on which address in the flow.json)

amyliu6: i see. So it's meant to achieve what `deploy.js` does too.  Maybe i missed it but how does it deploy all contracts? which part of code does that?

KsenijaZivkovic: Overflow library does it with this part of the code:
`&OverflowUtils{O: overflow.NewOverflowEmulator().Start()}`
it reads the flow.json and deploys all the contracts from there, and then we call the DeployFiatTokenContract function to deploy the FiatToken.cdc

jacksonConrad (approved):
looks good!

amyliu6 (approved):

PR #60, opened 2022-08-04
VSL-115: Reverted keys weight to 999 to enable Blocto signing

KsenijaZivkovic:
## Description

Lowered from 1000 to 999 total weight of the keys so that signing the message from one user works with Blocto wallet
More context on this thread: https://brud.slack.com/archives/C03ERT4F2MB/p1659551174397049 

## Demo / Test Result
Nothing to be shown, should fix the error when the contracts are deployed

### DIFF
--- a/contracts/DAOTreasury.cdc
+++ b/contracts/DAOTreasury.cdc
@@ -187,8 +187,8 @@ pub contract DAOTreasuryV2 {
         message: "Invalid Signature"
       )
       assert(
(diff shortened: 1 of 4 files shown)


amyliu6 (approved):
Still not 100% sure if it's ok to leave with it as it was originally reported as a security issue from Quantstamp. Will try to get more clarity on this.
But I think it's ok to make this change for now to at least unblock transaction on testnet.

jacksonConrad (approved):

PR #64, opened 2022-08-10
VSL-51: Added new asset flow - BE

KsenijaZivkovic:
## Description

- This PR is about adding assets over safe settings. These assets can be vaults or collections.
- In case of vaults, user needs to provide the name of the contract (FlowToken, FiatToken etc)
- In case of collections, user needs to provide both the name and the address of the contract.
- These values will be used to trigger the correct transaction and add vault/collection to the treasury.
- Added small piece of UI for enabling users to add assets.
- Added other tokens for testing on staging - Starly, Blocto and REVV

## Demo / Test Result
Safe Settings, below the Owners section, two buttons are added:
Screen Shot 2022-08-09 at 6 01 15 PM
### Add Vault Modal:
Screen Shot 2022-08-10 at 4 21 26 PM
Screen Shot 2022-08-10 at 4 21 36 PM
### Add Collection Modal:
Screen Shot 2022-08-09 at 6 02 16 PM

For now only add collection can be tested by using ExampleNFT contract on local testing environment, or on staging environment - any NFT contract deployed to testnet.
Adding vaults will be testable once we update flow.json with the names of the contracts and addresses of the accounts they are deployed, similar to how it is done in the next snapshot:
Screen Shot 2022-08-09 at 6 06 24 PM


amyliu6:
@KsenijaZivkovic , after user added Vault, it needs to be rendered under coin type dropdown: https://github.com/DapperCollectives/VESSEL/blob/5efbfc90581a863ce7eb99989b90d2ec199e414e/packages/client/src/components/SendTokens/components/CoinTypeDropDown.js#L9, can that be covered in this PR or we create another ticket for that work?

KsenijaZivkovic:
> @KsenijaZivkovic , after user added Vault, it needs to be rendered under coin type dropdown:
> 
> https://github.com/DapperCollectives/VESSEL/blob/5efbfc90581a863ce7eb99989b90d2ec199e414e/packages/client/src/components/SendTokens/components/CoinTypeDropDown.js#L9
> 
> , can that be covered in this PR or we create another ticket for that work?

Changed from text input to dropdown!

packages/client/src/constants/maps.js
```diff
@@ -10,20 +10,41 @@ export const ASSET_TYPE_TO_META = {
 
 export const COIN_TYPE_TO_META = {
   [COIN_TYPES.FLOW]: {
+    contract: "FlowToken",
```
amyliu6, line 13: can we rename the key to `contractName`?

packages/client/src/components/SafeSettings.js
```diff
@@ -349,6 +351,144 @@ const AddSafeOwner = ({ web3, onCancel, onNext, safeOwners }) => {
...
+    "button flex-1 p-4",
+    isFormValid ? "is-link" : "is-light is-disabled",
+  ];
+
+  const coinTypes = Object.entries(COIN_TYPE_TO_META).map((type) => ({
+    itemValue: type[0],
+    displayText: type[1].displayName,
+  }));
+
+  console.log(coinTypes);
```
linxiao-li, line 374: nit, we can remove this console log

packages/client/src/hooks/useTreasury.js
```diff
@@ -68,11 +70,9 @@ const doProposeTransfer = async (
   coinType
 ) => {
   const uFixAmount = String(parseFloat(amount).toFixed(8));
-  const tokenAddress =
-    process.env.REACT_APP_FLOW_ENV === "emulator"
-      ? "ee82856bf20e2aa6"
-      : "9a0766d93b6608b7";
-  const recepientVault = `Capability<&AnyResource{A.${tokenAddress}.FungibleToken.Receiver}>`;
+  const contractAddress = await config().get(`0x${COIN_TYPE_TO_META[coinType].contract}`);
```
amyliu6, line 73: Can we consider add a safe check for `COIN_TYPE_TO_META[coinType]`? I think it's unlikely to happen in current workflow but it's a good practice to error out explicitly in case input coinType is not one of the supported coin.

packages/client/src/components/SafeSettings.js
```diff
@@ -349,6 +351,142 @@ const AddSafeOwner = ({ web3, onCancel, onNext, safeOwners }) => {
...
+  ];
+
+  const coinTypes = Object.entries(COIN_TYPE_TO_META).map((type) => ({
+    itemValue: type[0],
+    displayText: type[1].displayName,
+  }));
+
+  const [coinType, setCoinType] = useState(coinTypes[0].displayText);
+
+  useEffect(() => {
```
linxiao-li, line 376: I don't see a user interface for the state contractName, guessing the purpose here is to project coingType to contractName. In this case can we pass the cointype instead of the contractName to the event handler (onAddVaultSubmit)? That way we can optimize out the extra state and useeffect rendering computation.

packages/client/src/flow/sendCollectionToTreasury.tx.js
```diff
@@ -8,11 +8,6 @@ export const SEND_COLLECTION_TO_TREASURY = `
 
 		prepare(signer: AuthAccount) {
 	
-			// borrow a reference to the signer's NFT collection
```
linxiao-li, line 11: just curious on the reason for removing this part

KsenijaZivkovic: It's not used anywhere?

packages/client/src/utils.js
```diff
@@ -46,39 +46,47 @@ export const isAddr = (addr) => {
...
+  }
+  if (isNFTTransfer) {
+    contractName = str.match(collectionRegex)[0].split(".")[2];
+    return str.replace(collectionRegex, contractName);
+  }
+  return str;
 };
 
+export const getTokenName = (contractName) => {
+  switch (contractName) {
```
amyliu6, line 72: Can we utilize COIN_TYPE_TO_META to get the tokenName instead? 
similar as this: https://github.com/DapperCollectives/VESSEL/blob/ba950e458a42315c4670a331be8f2dfeea2eeff7/packages/client/src/components/SendTokens/components/CoinTypeDropDown.js#L9-L12

packages/client/src/utils.js
```diff
@@ -46,39 +46,47 @@ export const isAddr = (addr) => {
...
+    return result.replace(vaultRegex, tokenName);
+  }
+  if (isNFTTransfer) {
+    contractName = str.match(collectionRegex)[0].split(".")[2];
+    return str.replace(collectionRegex, contractName);
+  }
+  return str;
 };
 
+export const getTokenName = (contractName) => {
```
linxiao-li, line 71: Suggest moving this map to constants/maps.js and name it `CONTRACT_NAME_TO_COIN_TYPE` to make the mapping relationship clear

packages/client/src/flow/sendCollectionToTreasury.tx.js
linxiao-li: Got it, I thought it was pre check for the collection's existence.

packages/client/src/utils.js
linxiao-li: @KsenijaZivkovic and I talked about this and we agreed to create another map, the `COIN_TYPE_TO_META` maps coinType to contractName, where in this use case we need the reversed mapping, sticking with `COIN_TYPE_TO_META` will require looping through the list of all coinTypes for each of the contract name we get, I think it's worthwhile to trade space for time for this instance.

linxiao-li (commented):
SafeSettings.js is almost 650 lines long, I think it's time for us to refactor it, created [this ticket](https://linear.app/dappercollectives/issue/VSL-138/refactor-safesettings-page) to track the work

packages/client/src/utils.js
amyliu6: i see 👍 . In this case support a new coin type will require changes at multiple places, maybe worth to add comment or something to help remind ppl these "connected" mappings. If we can have unit test to cover this mapping would be very helpful to serve as a safe guard.

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Will update, thank you!

packages/client/src/flow/sendCollectionToTreasury.tx.js
KsenijaZivkovic: It maybe was before, but now it seems the user's collection is not relevant anymore because it's ignored and a new collection is created

packages/client/src/flow/sendCollectionToTreasury.tx.js
```diff
@@ -8,11 +8,6 @@ export const SEND_COLLECTION_TO_TREASURY = `
 
 		prepare(signer: AuthAccount) {
```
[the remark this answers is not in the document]
KsenijaZivkovic: I'm wondering if we should keep this transaction, because now that we're adding new collections through add asset flow this is too covered. Then we will be able to remove the Deposit Collection functionality from the UI which was used just for testing. What do you think?

packages/client/src/hooks/useNFTs.js
```diff
@@ -46,12 +46,12 @@ const doSendCollectionToTreasury = async (
 const doProposeNFTTransfer = async (
```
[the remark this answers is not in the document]
KsenijaZivkovic: It should be generic enough to support all NFT collections

packages/client/src/hooks/useTreasury.js
KsenijaZivkovic: It can happen if the user tampers with the html, and those errors are their fault 😄 
Is it enough to add an error in the console or we would need to present an error on the UI?

packages/client/src/utils.js
KsenijaZivkovic: It would be ideal to add an API integration that would know how to map this correctly, @jacksonConrad 
 sent me this https://docs.flowgraph.co/examples/contract/token/
 Maybe for now to leave it hardcoded (add comment) and plan the integration for the future

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Updated!

packages/client/src/hooks/useNFTs.js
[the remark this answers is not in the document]
KsenijaZivkovic: Updated to be generic

packages/client/src/flow/sendCollectionToTreasury.tx.js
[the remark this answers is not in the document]
amyliu6: yes! since now we have `add_collection` which covers this functionality already, this one can be removed.

packages/client/src/hooks/useTreasury.js
amyliu6: i think a console log is good enough for now.

packages/client/src/flow/sendCollectionToTreasury.tx.js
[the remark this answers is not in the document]
KsenijaZivkovic: Removed this transaction and button which reuses it

packages/client/src/hooks/useTreasury.js
KsenijaZivkovic: Added error handling as part of proposeAddVault and proposeAddCollection -> error will be logged to the console error in the browser

packages/client/src/components/SafeSettings.js
```diff
@@ -349,6 +351,134 @@ const AddSafeOwner = ({ web3, onCancel, onNext, safeOwners }) => {
...
+    "button flex-1 p-4",
+    "is-link"
+  ];
+
+  const coinTypes = Object.entries(COIN_TYPE_TO_META).map((type) => ({
+    itemValue: type[0],
+    displayText: type[1].displayName,
+  }));
+
+  const [coinType, setCoinType] = useState(coinTypes[0].displayText);
```
linxiao-li, line 371: we should use itemvalue instead, they are the keys of the map.

packages/client/src/components/TestToolBox.js
```diff
@@ -62,9 +32,11 @@ const TestToolBox = ({ address }) => {
     setUserFUSDBalance(fusdBalance);
     setUserFlowBalance(flowBalance);
   };
+
```
linxiao-li, line 35: useEffect defined without dependency array will cause the callback function to run on each rendering. The purpose of the empty array here is to only run it once when component mounts. Since we have the refresh button, I'd say let's keep the empty dependency array to avoid unnecessary data fetching on most rendering cycles.

KsenijaZivkovic: If I remove it there is a warning from the compiler: React Hook useEffect has a missing dependency: 'updateUserBalance'. Either include it or remove the dependency array. 
Should we ignore the warning?

linxiao-li:
```suggestion
  useEffect(() => {
    updateUserBalance();
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```
we can disable eslint for that line, it will make the warning go away

contracts/DAOTreasury.cdc
```diff
@@ -206,8 +206,12 @@ pub contract DAOTreasuryV2 {
     pub fun depositTokens(identifier: String, vault: @FungibleToken.Vault) {
       emit DepositTokens(identifier: identifier)
 
-      let vaultRef = (&self.vaults[identifier] as &FungibleToken.Vault?)!
-      vaultRef.deposit(from: <- vault)
+      if(self.vaults[identifier] == nil) {
+        self.vaults[identifier] <-! vault
```
jacksonConrad, line 210: i think we do not want to let any user deposit a new vault if one doesnt exist for that identifier.  that is an action that only signers should be able to take. i thought we had a `signerDepositVault` function, similar to `signerDepositCollection` but looks like we don't.

i think we should revert this change so that this function will error out if the vault doesnt exist.
created ticket to track [here](https://linear.app/dappercollectives/issue/VSL-139/implement-signerdepositvault-in-daotreasurycdc)

jacksonConrad (changes requested):

packages/client/src/components/SafeSettings.js
KsenijaZivkovic: Why? They are the same values currently, is there a plan to change the display names in the future?

contracts/DAOTreasury.cdc
KsenijaZivkovic: So in the end we would need two functions -> one that allows everyone to deposit if the vault exists
And the other one -> where only the signer can add a new vault if it's not existing?

KsenijaZivkovic: If i just revert it it wouldn't be possible to test adding vaults to treasury - either i add signers here or merge an incomplete feature.
I tried adding message signature payload here but it's not feasible because this function is used in `TransferTokenToTreasury -> execute `
and it didn't make sense to add the signatureMessagePayload here because only the signers can execute actions, which would be a duplicate check

jacksonConrad: yes exactly, which i believe is how its working for NFT Collections

jacksonConrad: NFTs: `depositNFT` and `signerDepositCollection`
FTs: `depositTokens` and `signerDepositVault`

KsenijaZivkovic: Updated, can you check again?
In order to add some test coverage for this I'd need to add a new fungible token type to the core contracts and add some tests for it, because all the fungible token vaults are already added through create treasury

KsenijaZivkovic: Added test

packages/client/src/components/SafeSettings.js
linxiao-li: Good question, the short answer is to keep the integrity of the coinType enum we defined. The pattern we practice here is "key(coinType)->attribute(displayName, vaultId...)", 
1. all the attributes (including the displayName) is subjected to change
2. all attributes should only be accessed by key
3. when passing arguments around we should avoid passing attributes and opt in passing by reference

There are a couple benefits
1. Single source of truth: displayName could be formatted/capitalized due to feature requirements and loose its integrity as a key to access other attributes of a coinType
2. Scalability: passing by reference allows the code that touches that argument not being impacted when we make changes to specific attributes

packages/client/src/constants/maps.js
```diff
@@ -49,3 +70,27 @@ export const SAFE_TYPES_TO_META = {
     displayText: "Other",
   },
 };
+
+/** 
+ * TODO: Remove once API for fetching the name of the token is implemented
+ * For now every time the fungible token is added to flow.json, the token name should be added here 
+ * (in case it is different from the contract name) 
+*/
+export const CONTRACT_NAME_TO_COIN_TYPE = (contractName) => {
```
amyliu6, line 79: here since it's a function, maybe `getCoinTypeFromContractName` make more sense?

packages/client/src/hooks/useTreasury.js
```diff
@@ -458,6 +493,28 @@ export default function useTreasury(treasuryAddr) {
     await tx(res).onceSealed();
     await refreshTreasury();
   };
+
+  const proposeAddVault = async (coinType) => {
+    const contractName = COIN_TYPE_TO_META[coinType].contractName;
```
amyliu6, line 498: can we move this line into `try` block too and check COIN_TYPE_TO_META[coinType] exist before accessing contractName? If it does's exist, we can log this coinType is not supported.

KsenijaZivkovic: I can move it to try block. However, the error is not possible to happen because coinType is selected through a dropdown, the only way user can get this error is if they tamper the html, and in that case if they break the app it's only in their browser

packages/client/src/constants/maps.js
KsenijaZivkovic: Changed it to a map to keep it as a constant
```javascript
export const CONTRACT_NAME_TO_COIN_TYPE = {
  "FlowToken": "FLOW",
  "FUSD": "FUSD",
  "FiatToken": "USDC",
  "BloctoToken": "BLT",
  "StarlyToken": "STARLY",
  "REVV": "REVV"
}
```

packages/client/src/hooks/useTreasury.js
amyliu6: yea that's right. Not a big problem anyways. (i was thinking more from a function perspective as any client could invoke this method, add safe check to explicitly error out is better than undefined runtime exception from javascript.)

packages/client/src/utils.js
```diff
@@ -46,33 +45,26 @@ export const isAddr = (addr) => {
...
+  const vaultRegex = /A\..*\.Vault/g;
+  const collectionRegex = /A\..*\.Collection/g;
+  const floatRegex = /[0-9]*[.][0-9]+/;
+  const isTokenTransfer = vaultRegex.test(str);
+  const isNFTTransfer = collectionRegex.test(str);
+  let contractName;
+  let result;
+  if (isTokenTransfer) {
+    contractName = str.match(vaultRegex)[0].split(".")[2];
+    const tokenName = CONTRACT_NAME_TO_COIN_TYPE(contractName);
```
amyliu6, line 59: `CONTRACT_NAME_TO_COIN_TYPE` is now defined as Object, this line probably won't work anymore, can you please double check?

amyliu6 (approved):
LGTM!

packages/client/src/constants/maps.js
```diff
@@ -49,3 +70,17 @@ export const SAFE_TYPES_TO_META = {
...
   },
 };
+
+/** 
+ * TODO: Remove once API for fetching the name of the token is implemented
+ * For now every time the fungible token is added to flow.json, the token name should be added here 
+ * (in case it is different from the contract name) 
+*/
+export const CONTRACT_NAME_TO_COIN_TYPE = {
+  "FlowToken": "FLOW",
```
linxiao-li, line 80: Let's use the enums defined above
```suggestion
  "FlowToken": COIN_TYPES.FLOW,
  "FUSD":COIN_TYPES.FUSD
  ...
```

KsenijaZivkovic: Updated!

linxiao-li: I don't see the updates, did you push up your change?

KsenijaZivkovic: https://github.com/DapperCollectives/VESSEL/pull/64/files#diff-784256b57bc20779b89d00fe1a77c943db2bd4e174e263610990518f2293b560R79-R85:~:text=%22FiatToken%22%3A-,COIN_TYPES,-.USDC%2C

linxiao-li (approved):

packages/client/src/utils.js
```diff
@@ -101,6 +93,7 @@ export const syncSafeOwnersWithSigners = (signers, safeOwners) => {
   return updatedOwners;
 };
 export const getVaultId = (identifiers, coinType) => {
+  console.log(identifiers)
```
jacksonConrad, line 96: can remove

jacksonConrad, line 96:
```suggestion
```

PR #69, opened 2022-08-12
hide remove link when there's only one owner on the list

linxiao-li:
## Description

As per the requirements of this ticket https://linear.app/dappercollectives/issue/VSL-92/prevent-remove-button-if-only-one-owner-fe-work, adds logic to prevent rendering remove button when there's only one owner in the list
## Demo / Test Result

https://user-images.githubusercontent.com/7423845/184389633-15092587-106c-497a-8fb5-3f80f00ed8c3.mov

### DIFF
--- a/packages/client/src/components/SafeSettings.js
+++ b/packages/client/src/components/SafeSettings.js
@@ -450,7 +450,7 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
     proposeRemoveSigner,
   } = web3;
   const verifiedSafeOwners = safeOwners.filter((so) => so.verified);
-  const onlyOneOwnerLeft = verifiedSafeOwners.length === 1;
+  const canRemoveSigner = verifiedSafeOwners.length > threshold;
   const onEditNameSubmit = (newName) => {
     modalContext.closeModal();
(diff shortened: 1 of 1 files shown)


packages/client/src/components/SafeSettings.js
```diff
@@ -450,6 +450,7 @@ function SafeSettings({ address, web3, name, threshold, safeOwners }) {
     proposeRemoveSigner,
   } = web3;
   const verifiedSafeOwners = safeOwners.filter((so) => so.verified);
+  const onlyOneOwnerLeft = verifiedSafeOwners.length === 1;
```
jacksonConrad, line 453: i think we should generalize this to hide the `remove signer` button if `safeOwners.length == threshold`.  Because the safe will never let you have less signers than the threshold, it would be most useful to disable this button when removing a signer is not an option.

linxiao-li: That's a very good point, thanks for the insight!

jacksonConrad (approved):

KsenijaZivkovic (approved):

jacksonConrad (approved):

PR #70, opened 2022-08-12
VSL-128: Signer Reject Action

jacksonConrad:
## Description

- initialize  all addresses on `accountsVerified` to `nil` instead of `false` so we can support rejections AND approvals
- implement `signerRejectTransaction` which sets `accountsVerified[address] = false`.  it also will automatically delete the action if the `(total number of rejections) > signers.length - threshold`.
- refactor MyMultiSig.cdc to have `signerApproveAction`/`signerRejectAction` functions on `Manager` resource.
  - in order to destroy the action if a signer rejects it and it has enough rejections to be deleted, i needed to move the `signerRejectAction` function from an instance method on the MultiSignAction  resource, to an instance method on the Manager resource.
- removes the  `MultiSignAction.totalVerified` param and adds getter fns to iterate over the accountsVerified to fetch total rejections and total approvals (`getTotalRejectedForAction`/`getTotalApprovedForAction`)
- Adds/updates tests for signer approvals and rejections

Misc. updates
- small optimization: updates `Manager.threshold` from `UInt64` to `Int`

## Demo / Test Result
- `make test`


transactions/create_treasury.cdc
```diff
@@ -3,7 +3,7 @@ import FlowToken from "../contracts/core/FlowToken.cdc"
 import FiatToken from "../contracts/core/FiatToken.cdc"
 import FUSD from "../contracts/core/FUSD.cdc"
 
-transaction(initialSigners: [Address], initialThreshold: UInt64) {
+transaction(initialSigners: [Address], initialThreshold: Int) {
```
KsenijaZivkovic, line 6: It's possible to create treasury with threshold 0

KsenijaZivkovic: Also posible to add negative values - when we use int instead of uint

transactions/signer_revoke.cdc
```diff
@@ -28,11 +29,10 @@ transaction(treasuryAddr: Address, actionUUID: UInt64, message: String, keyIds:
 
   }
   execute {
-    self.isValid = self.action.signerRevokeApproval(messageSignaturePayload: self.messageSignaturePayload)
+    self.isValid = self.manager.signerRejectAction(actionUUID: actionUUID, messageSignaturePayload: self.messageSignaturePayload)
   }
 
   post {
     self.isValid == true: "Unable to revoke approval: invalid message or signature"
-    self.action.accountsVerified[self.messageSignaturePayload.signingAddr] == false: "Error: tx completed but signer approval not revoked"
```
KsenijaZivkovic, line 36: maybe we should leave this check too like for approve

contracts/MyMultiSig.cdc
```diff
@@ -252,6 +151,86 @@ pub contract MyMultiSigV2 {
...
+            let action = self.borrowAction(actionUUID: actionUUID)
+
+
+            // Revoke approval
+            action.setAccountsVerified(signer: messageSignaturePayload.signingAddr, value: false)
+
+            emit ActionRejectedBySigner(address: messageSignaturePayload.signingAddr, uuid: self.uuid)
+
+            // Check if we have enough rejections to delete the action
+            if  self.getTotalRejectedForAction(actionUUID: actionUUID) > (self.signers.length - Int(self.threshold)) {
```
KsenijaZivkovic, line 219: nit: brackets in the if clause

contracts/MyMultiSig.cdc
```diff
@@ -252,6 +151,86 @@ pub contract MyMultiSigV2 {
...
+            // Validate Signature
+            var signatureValidationResponse = MyMultiSigV2.validateSignature(payload: messageSignaturePayload)
+            assert(signatureValidationResponse.isValid == true, message: "Invalid Signatures")
+            assert(signatureValidationResponse.totalWeight >= 999.0, message: "Total weight of combined signatures did not satisfy 999 key weight requirement.")
+
+            // Approve action
+            action.setAccountsVerified(signer: messageSignaturePayload.signingAddr, value: true)
+
+            emit ActionApprovedBySigner(address: messageSignaturePayload.signingAddr, uuid: self.uuid)
+            return signatureValidationResponse.isValid
```
KsenijaZivkovic, line 186: This will always be true, because of the assertion above, so I think it's safe to remove the return type for both signerApproveAction and signerRejectAction

transactions/create_treasury.cdc
jacksonConrad: ahh true, i'll just use a regular `UInt` then. instead of 64 bit uint

transactions/signer_revoke.cdc
jacksonConrad: the reason I removed it is because the case where the signer revokes it, and the action gets destroyed.  in this case, the transaction always fails on this `post` condition because `self.action` no longer exists.

i couldn't figure out any way to check if `self.action` has been destroyed. and something like `self.action && self.action.accountsVerified[addr] == false` also fails, because it is still trying to evaluate `self.action.accountsVerified`

I'll ask in AZ slack to confirm if its possible to check whether a resource you have a reference to has been destroyed.

main_test.go
```diff
@@ -473,12 +473,64 @@ func TestTransferNonFungibleTokensToTreasuryActions(t *testing.T) {
...
+
+		transferTokenActionUUID = keys[0]
+
+		// All but one signer rejects approval so we dont delete the action
+		for i, signer := range Signers {
+			if i == len(Signers)-1 {
+				continue
+			}
+			otu.SignerRejectApproval("treasuryOwner", transferTokenActionUUID, signer)
+		}
```
amyliu6, line 505: shall we **assert** here that the action is still there?

contracts/MyMultiSig.cdc
```diff
@@ -272,19 +247,37 @@ pub contract MyMultiSigV2 {
             return self.signers
         }
 
-        pub fun getThreshold(): UInt64 {
+        pub fun getThreshold(): UInt {
             return self.threshold
         }
 
         pub fun getVerifiedSignersForAction(actionUUID: UInt64): {Address: Bool} {
             return self.borrowAction(actionUUID: actionUUID).accountsVerified
```
amyliu6, line 255: i'm a bit confused with this naming now. 😅 
`accountsVerified` could be true/false/nil. `getVerifiedSignersForAction` makes me think this is to get all accounts that approved the action, but we are actually just getting all signers, right? shall we name it `getAllSignersForAction` ?

Same on FE, we are storing this data under https://github.com/DapperCollectives/VESSEL/blob/183db5f8733bbe69356a5518c0ff1f10cf7ab50c/packages/client/src/hooks/useTreasury.js#L293

Here instead of `verifiedSigners`, can we consider name it `allSigners`?

packages/client/src/flow/updateThreshold.tx.js
```diff
@@ -3,7 +3,7 @@ export const UPDATE_THRESHOLD = `
   import TreasuryActionsV2 from 0xTreasuryActionsV2
   import MyMultiSigV2 from 0xMyMultiSigV2
 
-  transaction(treasuryAddr: Address, newThreshold: UInt64, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
+  transaction(treasuryAddr: Address, newThreshold: Int, message: String, keyIds: [UInt64], signatures: [String], signatureBlock: UInt64) {
```
KsenijaZivkovic, line 6: UInt here?

packages/client/src/flow/initializeTreasury.tx.js
```diff
@@ -4,7 +4,7 @@ export const INITIALIZE_TREASURY = `
 	import FiatToken from 0xFiatToken
 	import FUSD from 0xFUSD
 	
-	transaction(initialSigners: [Address], initialThreshold: UInt64) {
+	transaction(initialSigners: [Address], initialThreshold: Int) {
```
KsenijaZivkovic, line 7: Same here, UInt

packages/client/src/hooks/useTreasury.js
```diff
@@ -149,7 +149,7 @@ const doUpdateThreshold = async (treasuryAddr, newThreshold) => {
     cadence: UPDATE_THRESHOLD,
     args: (arg, t) => [
       arg(treasuryAddr, t.Address),
-      arg(newThreshold, t.UInt64),
+      arg(newThreshold, t.Int),
```
KsenijaZivkovic, line 152: t.UInt on all places 😄

contracts/MyMultiSig.cdc
jacksonConrad: @amyliu6 yes, I think thats a much better naming convention! i'll make that update.

contracts/MyMultiSig.cdc
```diff
@@ -263,6 +152,89 @@ pub contract MyMultiSigV2 {
...
+        pub fun signerApproveAction(actionUUID: UInt64, messageSignaturePayload: MessageSignaturePayload) {
+            pre {
+                self.signers[messageSignaturePayload.signingAddr] == true:
+                    "This address is not allowed to sign for this."
+                self.borrowAction(actionUUID:actionUUID).accountsVerified[messageSignaturePayload.signingAddr] != true:
+                    "This address has already signed."
+            }
+            let action = self.borrowAction(actionUUID:actionUUID)
+
+            // Validate Message
```
linxiao-li, line 172: We are validating message signature for both approving action and rejecting action, is it possible to abstract this process into a private function and have the two public ones call it?

contracts/MyMultiSig.cdc
jacksonConrad: @amyliu6 I've updated the naming of `accountsVerified`/`verifiedSigners` to `signerResponses` and `getApprovedSignersForAction`/`getRejectedSignersForAction`.  In general I think we should move away from the term `verified` to describe signer status... it should be `approved`/`rejected`/`pending`.

i've also updated `signerResponses` type from `{Address: Bool}` to `{Address: SignerResponse}` where `SignerResponse` is an `enum` type.  The issue with using `Bool` is that it seems in Golang, if you try to access a `map[string]bool` where the key doesnt exist, you get `false` instead of `nil`, so it wasn't working as cleanly as I was hoping.  Additionally, this gives us more flexibility down the road to add more SignerResponse statuses (i.e. `abstain` is an option that similar products support).

contracts/MyMultiSig.cdc
jacksonConrad: we are already sort of doing this -- both `signerApprove` and `signerReject` functions have:

```
// Validate Message
assert(
       MyMultiSigV2.approveOrRejectActionMessageIsValid(action: action, messageSignaturePayload: messageSignaturePayload),
        message: "Signed message is invalid"
)
```

transactions/signer_revoke.cdc
jacksonConrad: it is not possible to check if a resource is destroyed in Cadence, according to Bastian.

we could add a separate return value which indicates that the resource has been destroyed, but i don't think its necessary.

contracts/MyMultiSig.cdc
jacksonConrad: @KsenijaZivkovic are you saying that it should be formatted like this?
```
if something
{
// stuff
}
```

jacksonConrad: i think i prefer the open bracket on the same line as the `if` clause, personally

amyliu6 (commented):
Looks great! with the new naming and enum it's much more readable now! 👍

KsenijaZivkovic (approved):

contracts/MyMultiSig.cdc
linxiao-li: Sorry about the incomplete message, what I mean is abstracting the process of `validateMessage` and `validateSignature` for both approving action and rejecting action

contracts/MyMultiSig.cdc
```diff
@@ -263,6 +157,89 @@ pub contract MyMultiSigV2 {
...
+
+            // Approve action
+            action.setSignerResponse(signer: messageSignaturePayload.signingAddr, value: SignerResponse.approved)
+
+            emit ActionApprovedBySigner(address: messageSignaturePayload.signingAddr, uuid: self.uuid)
+        }
+
+        pub fun signerRejectAction(actionUUID: UInt64, messageSignaturePayload: MessageSignaturePayload) {
+            pre {
+                self.actions[actionUUID] != nil: "Couldn't find action with UUID ".concat(actionUUID.toString())
```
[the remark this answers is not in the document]
jacksonConrad: good catch

contracts/MyMultiSig.cdc
jacksonConrad: Gotcha!!! Its a good suggestion!  The reason I haven't done that is because I've been anticipating that we might eventually change the message format for these two functions, so the validations would be slightly different.  I've kept them the same for now just to make it easy to implement and test, but I think we need to have a broader conversation on what the contents of the message actually should be for each function to make it as secure as possible.

To summarize -- having identical message for approving & rejecting has some security issues, and I plan on changing it. :)

contracts/MyMultiSig.cdc
```diff
@@ -283,19 +260,47 @@ pub contract MyMultiSigV2 {
...
+            return total
         }
 
-        pub fun getTotalVerifiedForAction(actionUUID: UInt64): UInt64 {
-            return self.borrowAction(actionUUID: actionUUID).totalVerified
+        pub fun getTotalRejectedForAction(actionUUID: UInt64): Int {
+            var total: Int = 0
+            let action = self.borrowAction(actionUUID: actionUUID)
+            for key in self.signers.keys {
+                if action.signerResponses[key] != nil && action.signerResponses[key]! == SignerResponse.rejected {
```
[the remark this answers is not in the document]
jacksonConrad: i'm not sure i understand this suggestion.. why would we need to check `value != approved` if we are checking `value == rejected`?

FYI I just  a slight re-write of this to make it more readable.

contracts/MyMultiSig.cdc
```diff
@@ -283,19 +260,47 @@ pub contract MyMultiSigV2 {
...
+                }
+            }
+            return  allResponses
+        }
+
+        pub fun getTotalApprovedForAction(actionUUID: UInt64): Int {
+            var total: Int = 0
+            let action = self.borrowAction(actionUUID: actionUUID)
+            for key in self.signers.keys {
+                if action.signerResponses[key] != nil && action.signerResponses[key]! == SignerResponse.approved {
```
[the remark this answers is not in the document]
jacksonConrad: i'm not sure i understand this suggestion.. why would we need to check `value != approved` if we are checking `value == rejected`?

FYI I just  a slight re-write of this to make it more readable.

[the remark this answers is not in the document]
amyliu6: the func is to `getTotalApprovedForAction`, shouldn't it look for "approved"? 
https://github.com/DapperCollectives/VESSEL/pull/70/files#diff-d6c67e336e7c8eb3ba7d49cd0a16be12910ac98006b41b6abfddf320a36295f1R284-R292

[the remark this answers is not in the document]
amyliu6: oh my bad, i read `! ==` as `!==`

amyliu6 (approved):

contracts/MyMultiSig.cdc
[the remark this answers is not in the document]
jacksonConrad: Ohh wait I think i see what you are saying, but I think the confusion is the `!` operator which is a "force" operator, that is not a `!==`!

```
if status != nil && status! == SignerResponse.approved
```

In this case `status` has type `SignerResponse?` which is an optional type. the `!` operator forces it to a non-optional, so we can compare its equality to `SignerResponse.approved`

PR #72, opened 2022-08-15
VSL-130: Added error modal for displaying errors on the UI

KsenijaZivkovic:
## Description

Added error modal for displaying errors on the UI.
Those errors are extracted from the cadence errors that contain some information that is not useful for the user, and this will have to be changed once the errors are structured.

## Demo / Test Result
Screen Shot 2022-08-15 at 4 14 05 PM
Screen Shot 2022-08-15 at 4 15 43 PM

Other errors:
Screen Shot 2022-08-15 at 4 15 03 PM
Screen Shot 2022-08-15 at 4 14 27 PM


packages/client/src/utils.js
```diff
@@ -106,3 +106,17 @@ export const getVaultId = (identifiers, coinType) => {
...
+
+export const showError = (modalContext, error) => {
+  modalContext.openModal(
+    <ErrorModal error={extractError(error)} />
+  );
+}
+
+function extractError(error) {
+  if (typeof error == "string" && error.includes("\n") && error.includes(": ")) {
+    const errorMessage = error.split("\n")[1];
```
KsenijaZivkovic, line 118: This part is hardcoded, and works for pre-conditions, post-conditions, panic and assert

amyliu6: I like your idea to wrap message with special character 😄 , that seems more reliable then grabbing it from ": " as per cadence team they might restructure their error object. But if we wrap message with @@ or anything like that, that could be our unique identifier to grab the message. What do you think?

KsenijaZivkovic: It's not gonna work for messages that come from core contracts unfortunately, I'm not sure how to handle those, since they cannot be wrapped in special characters. What do you think how can we handle those errors?

amyliu6: hmm I thought majority of the error use case that we want to surface is around our own treasury contract only. What kind of message are returning from core contracts? any examples?

KsenijaZivkovic: There are some examples in the PR description - when sending fungible tokens or NFT fail then the errors from those contracts can happen

amyliu6: i see... in that case can we work with Sara to add a Cypress test to cover at least one UI message as part of regression suite? In case anything changed we can catch it during regression testing.

KsenijaZivkovic: Sounds great! And once we update the libraries in case something changes we will catch this right away and update it!

amyliu6: Created a ticket here: https://linear.app/dappercollectives/issue/VSL-143/add-test-to-cover-error-scenario-in-regression-suite

amyliu6 (approved):

packages/client/src/App.sass
```diff
@@ -24,11 +24,13 @@ $input-focus-border-color: transparent
 $input-border-color: black
 $input-shadow: none
 $button-background-color: transparent
-$button-border-color: black
+$button-border-color: #7464F0
```
linxiao-li, line 27: I believe this changes the button styles globally, is this to align with the new design?

packages/client/src/utils.js
```diff
@@ -106,3 +106,17 @@ export const getVaultId = (identifiers, coinType) => {
     (id) => id.indexOf(COIN_TYPE_TO_META[coinType].vaultName) >= 0
   );
 };
+
+export const showError = (modalContext, error) => {
```
linxiao-li, line 110: Let's encapsulate the showError and extractError into a `useErrorMessage` hook. Utils should not be aware of the components

packages/client/src/App.sass
KsenijaZivkovic: Yes!

packages/client/src/utils.js
KsenijaZivkovic: Updated, tnx for help!

linxiao-li (approved):

PR #73, opened 2022-08-16
Updated TestToolBox to avoid exception

amyliu6:
## Description

**Issue**:
if test account does not have FUSD vault, deposit token will throw exception:
Screen Shot 2022-08-16 at 11 44 38 AM

In this PR moved FUSD setup as optional step, so that if user doesn't have FUSD vault, FLOW tokens can still be deposited successfully.

## Demo / Test Result

After this change, deposit token will work with no issues and error msg on depositing FUSD will be logged in console:
Screen Shot 2022-08-16 at 11 50 42 AM


KsenijaZivkovic (approved):

linxiao-li (approved):
LGTM

PR #74, opened 2022-08-17
VSL-142: Enable signing with Lilico and Blocto

KsenijaZivkovic:
## Description

Enabling signing with Lilico and Blocto
For Lilico the problem was verifying the hashing algorithm which is different than Blocto
For Blocto the problem was that on testnet it started returning both keys for signing the messages - temporary solution is to use the one with 999 weight, until we research how to send and validate both

## Demo / Test Result
`make test`

### DIFF
--- a/contracts/MyMultiSig.cdc
+++ b/contracts/MyMultiSig.cdc
@@ -361,9 +361,9 @@ pub contract MyMultiSigV2 {
             keyList.add(
                 PublicKey(
                     publicKey: accountKey.publicKey.publicKey,
-                    signatureAlgorithm: UInt(accountKey.publicKey.signatureAlgorithm.rawValue) == 2 ? SignatureAlgorithm.ECDSA_secp256k1  : SignatureAlgorithm.ECDSA_P256
+                    signatureAlgorithm: accountKey.publicKey.signatureAlgorithm
                 ),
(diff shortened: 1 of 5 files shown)


contracts/MyMultiSig.cdc
```diff
@@ -363,7 +363,7 @@ pub contract MyMultiSigV2 {
                     publicKey: accountKey.publicKey.publicKey,
                     signatureAlgorithm: UInt(accountKey.publicKey.signatureAlgorithm.rawValue) == 2 ? SignatureAlgorithm.ECDSA_secp256k1  : SignatureAlgorithm.ECDSA_P256
                 ),
-                hashAlgorithm: HashAlgorithm.SHA3_256,
+                hashAlgorithm: UInt(accountKey.hashAlgorithm.rawValue) == 3 ? HashAlgorithm.SHA3_256 : HashAlgorithm.SHA2_256,
```
amyliu6, line 366: curious what value do we get from this: `accountKey.hashAlgorithm.rawValue` ?

KsenijaZivkovic: 1 or 3 :D

KsenijaZivkovic: Updated for both hashing and signature algorithms!

amyliu6 (approved):

jacksonConrad (approved):

PR #75, opened 2022-08-18
VSL 138: Refactor Safe Settings Page

linxiao-li:
## Description

This pr refactors the safe settings page and fixes the [bug of incorrect signature number](https://linear.app/dappercollectives/issue/VSL-140/threshold-updated-before-executing-action) 

## Demo / Test Result
Edit safe name

https://user-images.githubusercontent.com/7423845/185295478-80f84ed2-f57b-45fd-ad25-39de0080eea8.mov

Increase threshold

https://user-images.githubusercontent.com/7423845/185295595-b14fe80e-e3f4-469e-80c2-c7db11c7b425.mov

Add signer

https://user-images.githubusercontent.com/7423845/185295686-68cb40e0-c2a2-48de-8ca3-619b3e0d4b99.mov

Add Signer and increase threshold

https://user-images.githubusercontent.com/7423845/185295768-87e91472-4157-4234-b86d-ee53fd54c3e7.mov

Remove signer

https://user-images.githubusercontent.com/7423845/185295797-61cd5a3e-244d-492b-84e7-691829bce828.mov


packages/client/src/components/SafeSettings/components/SafeDetailsSetting/EditSafeName.js
```diff
@@ -0,0 +1,50 @@
+import { useState } from "react";
+
+const EditSafeName = ({ name, onCancel, onSubmit }) => {
+  const [currSafeName, setCurrSafeName] = useState(name);
```
KsenijaZivkovic, line 4: nit: we can rename to safeName

packages/client/src/components/SafeSettings/index.js
```diff
@@ -0,0 +1,22 @@
+import React, { useContext } from "react";
+import { useLocation } from "react-router-dom";
+import { Web3Context } from "contexts/Web3";
+import SafeDetailsSetting from "./components/SafeDetailsSetting";
+import SignatureThreshold from "./components/SignatureThreshold";
+import Owners from "./components/Owners";
+
+const SafeSettings = () => {
```
KsenijaZivkovic, line 8: Can we add treasuryAddress as input param for SafeSettings instead of fetching it from the path?

packages/client/src/components/SafeSettings/components/SafeDetailsSetting/EditSafeName.js
linxiao-li: I actually think we should keep the edit part, just to differentiate it from other components that only displays safe name, like the one on home screen

packages/client/src/components/SafeSettings/index.js
linxiao-li: This is to decouple the safeSettings component from the Safe page and to avoid property drill

amyliu6 (approved):
👏  LGTM!

packages/client/src/components/SafeSettings/components/SafeDetailsSetting/EditSafeName.js
KsenijaZivkovic: currSafeName -> safeName

packages/client/src/components/SafeSettings/index.js
KsenijaZivkovic: I've read a bit about prop drilling, "Prop drilling is basically a situation when the same data is being sent at almost every level due to requirements in the final level."
Here we would just pass this address once to this SafeSettings component, which is not that bad
Safe.js already contains fetching the address from params:
```
  const params = useParams();
  const { address, tab } = params;
```
I'm concerned about this line, and if url on staging or prod is slightly different we might encounter errors:
```
const treasuryAddress = location.pathname.split("/")[2];
```

packages/client/src/components/SafeSettings/components/SafeDetailsSetting/EditSafeName.js
linxiao-li: lol I see, I thought you were talking about the file name, my bad, yes good idea!

packages/client/src/components/SafeSettings/index.js
linxiao-li: You are right, passing an address between those two components won't be a property drill, but it still creates dependency. The higher order of the component, the less dependent it should be with its parents, and comparing with having the key information passed down the chain,  it's actually more consistent and reliable using the pathname, the order is always gonna be what we defined in the code, refer[ this doc](https://v5.reactrouter.com/web/api/location) for the structure of location

KsenijaZivkovic (approved):

KsenijaZivkovic:
Now I've realized that with this PR the UI for Add Vault and Add Collection is removed, oops, we should revert this or re-add it @linxiao-li

linxiao-li:
@KsenijaZivkovic Oh no... I'll open up another pr to add them, do you remember which pr they originally belong?

KsenijaZivkovic:
> @KsenijaZivkovic Oh no... I'll open up another pr to add them, do you remember which pr they originally belong?

Here it is, as part of SafeSettings.js https://github.com/DapperCollectives/VESSEL/pull/64

PR #76, opened 2022-08-18
VSL-142: Added FCLCrypto contract and reused it's validation

KsenijaZivkovic:
## Description

Added FCLCrypto.cdc to core contracts and reused it's function for validating signature
Removed MyMultiSig validateSignature function
Updated UI transactions to send also the messageHex which is used for validating signatures, beside parameter called message which is used to validate the content of the message

## Demo / Test Result

`make test`


contracts/MyMultiSig.cdc
```diff
@@ -39,13 +40,15 @@ pub contract MyMultiSigV2 {
     pub struct MessageSignaturePayload {
         pub let signingAddr: Address
         pub let message: String
+        pub let messageHex: String
         pub let keyIds: [Int]
         pub let signatures: [String]
         pub let signatureBlock: UInt64
 
-        init(signingAddr: Address, message: String, keyIds: [Int], signatures: [String], signatureBlock: UInt64) {
+        init(signingAddr: Address, message: String, messageHex: String, keyIds: [Int], signatures: [String], signatureBlock: UInt64) {
```
amyliu6, line 48: instead of passing in Hex string, can we convert string to uft8, then encode to Hex?
`cadence•let utf8: [UInt8]`
`String.encodeHex(data)` (ref: https://developers.flow.com/cadence/language/values-and-types)

KsenijaZivkovic:
Closing because not all commits are reflected

PR #79, opened 2022-08-22
vsl 99: Flowfee estimation on send modal

linxiao-li:
## Description

This pr adds the support for showing the estimated transaction fee before sending tokens from treasury to account as per request of [this ticket](https://linear.app/dappercollectives/issue/VSL-99)

## Demo / Test Result
**The estimation will always show 0 on emulator, connect to testnet to see the real result**
1. in `packages/client/.env` change `REACT_APP_FLOW_ENV=emulator` to `REACT_APP_FLOW_ENV=testnet`
2. in root directory, run `yarn update-contracts` 
3. follow the steps to send any amount of tokens to a valid address on testnet
4. on Review screen, you should see
Screen Shot 2022-08-22 at 8 35 22 AM


contracts/core/FlowFees.cdc
```diff
@@ -0,0 +1,131 @@
+import FungibleToken from "./FungibleToken.cdc"
```
KsenijaZivkovic, line 1: Add a comment including repository and commit for this contract

packages/client/src/hooks/useFlowFees.js
```diff
@@ -0,0 +1,24 @@
...
+import { GET_FLOW_FEES_ESTIMATION } from "../flow";
+import {
+  EXECUTION_EFFORTS,
+  DEFAULT_INCLUSION_EFFORT,
+} from "constants/constantValues";
+
+const useFlowFees = () => {
+  const getEstimation = async () => {
+    const { PROPOSE_ACTION, SIGN_ACTION, EXECUTE_ACTION } = EXECUTION_EFFORTS;
+    const executionEffort = PROPOSE_ACTION + SIGN_ACTION + EXECUTE_ACTION;
```
KsenijaZivkovic, line 11: I'm not sure if we should just sum these three up, because we are having 3 separate actions here, and each should have its own estimation?

contracts/core/FlowFees.cdc
linxiao-li: Could you elaborate?

KsenijaZivkovic: All copied contracts should have a reference to the repository where they came from, for example:
https://github.com/DapperCollectives/VESSEL/blob/main/contracts/core/BloctoToken.cdc#L3

packages/client/src/hooks/useFlowFees.js
linxiao-li: base on the design, I made the assumption that it's a sum, watching the [conversation here](https://brud.slack.com/archives/C03ERT4F2MB/p1661187075427139) for the direction

jacksonConrad: Yeah, I don't think it should be summed.  Instead, `getEstimation` should accept tx type as an argument, and `executionEffort` should be assigned based on that.

jacksonConrad (changes requested):

amyliu6 (approved):

linxiao-li:
@jacksonConrad That's a good idea, do we have a enum list for actions in the backend?

jacksonConrad (approved):

jacksonConrad:
> @jacksonConrad That's a good idea, do we have a enum list for actions in the backend?

sorry what was this in response to?

PR #80, opened 2022-08-22
VSL-142: signature validation

KsenijaZivkovic:
## Description

Added FCLCrypto.cdc to core contracts and reused it's function for validating signature
Removed MyMultiSig validateSignature function
Updated init method for creating message signature payload to include also messageHex, which is needed for validating the signatures

## Demo / Test Result

`make test`


contracts/DAOTreasury.cdc
```diff
@@ -95,16 +96,17 @@ pub contract DAOTreasuryV2 {
...
+        keyIndices: signaturePayload.keyIds,
+        signatures: signaturePayload.signatures
+      )
 
       assert(
-        signatureValidationResponse.isValid == true,
+        signatureValidationResponse == true,
         message: "Invalid Signature"
       )
-      assert(
```
KsenijaZivkovic, line 104: This is included in FCLCrypto - check if the weight is less than 1000

jacksonConrad (approved):

amyliu6 (approved):
LGTM!

PR #82, opened 2022-08-24
move FiatToken deployment logic to deploy.js

mannynotfound:
Currently we cannot deploy `contracts/core/FiatToken.cdc` on windows machines because the `overflow` package does not support it. Additionally, we do not need to rely on `overflow` when we already had a `deploy.js` script that is now dead weight. This removes the overflow deployment script and adjusts `deploy.js` and `deploy_contract_with_auth.cdc` to work nicely together.


KsenijaZivkovic (changes requested):
Revert deleting test_utils.go, and remove this line and import of util
```
util.DeployFiatToken("treasuryOwner", "USDC", "0.1.0", otu.O)
```

mannynotfound:
@KsenijaZivkovic test_utils.go should be right now

KsenijaZivkovic (approved):

jacksonConrad (approved):

PR #84, opened 2022-08-24
VSL-150: Send USDC token to account

linxiao-li:
## Description
This pr implemented workflow to send USDC token from a treasury to an account

- added `getUserUSDCBalance` script
- created `useAccount` custom hook to handle account related mutations and queries, moved `getUserBalance` and `initDepositTokensToTreasury` from useTreasury to `useAccount`  
- created `constants` file to store gas limit constants, moved limit constants away from web3 context

## Demo / Test Result
** no extra set up required to seed the emulator account with USDC tokens**

https://user-images.githubusercontent.com/7423845/186504632-bed37a58-c3c5-4132-9c54-60eff97052fd.mov


packages/client/src/components/TestToolBox.js
```diff
@@ -1,39 +1,33 @@
...
 import { Web3Context } from "contexts/Web3";
 import { COIN_TYPES } from "constants/enums";
+import { COIN_TYPE_TO_META } from "constants/maps";
+import { useAccount } from "hooks";
+import { formatAddress } from "utils";
 const TestToolBox = ({ address }) => {
   const [showToolBox, setShowToolBox] = useState(false);
-  const [userFUSDBalance, setUserFUSDBalance] = useState(0);
-  const [userFlowBalance, setUserFlowBalance] = useState(0);
+  const [userBalances, setuserBalances] = useState([]);
```
KsenijaZivkovic, line 9: nit: setUserBalances

packages/client/src/components/TestToolBox.js
```diff
@@ -60,23 +55,35 @@ const TestToolBox = ({ address }) => {
             <div className="mr-5">
               <label>
                 <strong>Balances</strong>
-                <a className="is-underlined ml-2" onClick={updateUserBalance}>
+                <button
+                  href="#"
```
KsenijaZivkovic, line 59: Do we need href on a button? Does this change its view to link?

packages/client/src/components/TestToolBox.js
```diff
@@ -60,23 +55,35 @@ const TestToolBox = ({ address }) => {
...
-                    <li>Flow: {userFlowBalance}</li>
-                    <li>FUSD: {userFUSDBalance}</li>
-                  </ul>
+                  {userBalances && !!userBalances.length && (
+                    <ul className="ml-3">
+                      {userBalances.map((coin) => (
+                        <li key={COIN_TYPE_TO_META[coin.coinType].displayName}>
+                          {`${COIN_TYPE_TO_META[coin.coinType].displayName}: ${
+                            coin.balance
+                          }`}
```
KsenijaZivkovic, line 75: Would this work too?
```
`${COIN_TYPE_TO_META[coin.coinType].displayName}: ${coin.balance}`
```
Note: the only change is without the curly braces around the expression

packages/client/src/components/TestToolBox.js
```diff
@@ -1,39 +1,33 @@
...
+
   const updateUserBalance = async () => {
     try {
-      const fusdBalance = await getUserFUSDBalance(user.addr);
-      setUserFUSDBalance(fusdBalance);
+      const newUserBalances = await getUserBalances(formatAddress(user.addr));
+      setuserBalances(newUserBalances);
     } catch (err) {
-      console.log(`Failed to get FUSD balance, error: ${err}`)
+      console.log(`Failed to get FUSD balance, error: ${err}`);
```
KsenijaZivkovic, line 29: Update error message, now we are catching for both FUSD and USDC errors

packages/client/src/hooks/useTreasury.js
```diff
@@ -220,14 +203,21 @@ const doProposeAddVault = async (treasuryAddr, contractName) => {
...
-  const intent = `A.${contractAddress.replace("0x", "")}.${contractName}.Collection`;
+const doProposeAddCollection = async (
+  treasuryAddr,
+  contractName,
+  contractAddress
+) => {
+  const intent = `A.${contractAddress.replace(
+    "0x",
+    ""
+  )}.${contractName}.Collection`;
```
amyliu6, line 220: is this change due to auto-formatting? It would be helpful to make sure we have same settings to avoid back and forth formatting changes in different commit. :)

packages/client/src/hooks/useTreasury.js
```diff
@@ -220,14 +203,21 @@ const doProposeAddVault = async (treasuryAddr, contractName) => {
...
 };
 
-const doProposeAddCollection = async (treasuryAddr, contractName, contractAddress) => {
-  const intent = `A.${contractAddress.replace("0x", "")}.${contractName}.Collection`;
+const doProposeAddCollection = async (
+  treasuryAddr,
+  contractName,
+  contractAddress
+) => {
+  const intent = `A.${contractAddress.replace(
```
KsenijaZivkovic, line 217: nit: add a util function that will do the replacing instead of address.replace("0x","") -> so we make it more readable for the future

packages/client/src/hooks/useAccount.js
```diff
@@ -0,0 +1,85 @@
...
+  SEND_TOKENS_TO_TREASURY,
+} from "flow";
+import { COIN_TYPE_TO_META } from "constants/maps";
+import { COIN_TYPES } from "constants/enums";
+import { REGULAR_LIMIT } from "constants/constants";
+
+const COIN_TYPE_LIST = [COIN_TYPES.FLOW, COIN_TYPES.FUSD, COIN_TYPES.USDC];
+
+export default function useAccount() {
+  const doSendTokensToTreasury = async (treasuryAddr, amount, coinType) => {
```
KsenijaZivkovic, line 15: I'm wondering if we should allow only signers to add tokens to the treasury?
@jacksonConrad @amyliu6 @linxiao-li 
If we decide to change the logic around this, we will tackle it in another task, this is just for discussion purposes only 😄

packages/client/src/hooks/useAccount.js
```diff
@@ -0,0 +1,85 @@
...
+      case COIN_TYPES.FLOW:
+        cadence = GET_USER_FLOW_BALANCE;
+        break;
+      case COIN_TYPES.FUSD:
+        cadence = GET_USER_FUSD_BALANCE;
+        break;
+      case COIN_TYPES.USDC:
+        cadence = GET_USER_USDC_BALANCE;
+      default:
+        cadence = GET_USER_FLOW_BALANCE;
```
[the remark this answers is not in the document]
jacksonConrad: +1

packages/client/src/hooks/useAccount.js
jacksonConrad: we've discussed this with Jose, and the intended functionality is that only signers can add a net-new `Vault` to the Treasury, but ANYONE can send tokens to an existing Vault in the treasury.

packages/client/src/components/TestToolBox.js
linxiao-li: Good catch! Will remove

packages/client/src/hooks/useAccount.js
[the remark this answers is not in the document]
linxiao-li: The only reason I'm hesitate to do that is, for each token type we want to retrieve balance we will need to import the token's contract (since not only the path is different, the vault type is different as well) and add to the conditions inside of the script (imagine moving the above switch statement into the script), that means this query script will get modified a lot. Feels like an anti-pattern for me. Not sure if we have a way to dynamically import contract in the script level?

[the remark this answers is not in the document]
amyliu6: i was thinking something like this: https://github.com/DapperCollectives/VESSEL/blob/c13812bd4c508cb97cf9323ca9e4453b4087b422/packages/client/src/flow/addCollection.tx.js#L1-L5, we can use token to fetch it's corresponding contractName and path by utilizing maps.js then pass into the template script
https://github.com/DapperCollectives/VESSEL/blob/c13812bd4c508cb97cf9323ca9e4453b4087b422/packages/client/src/constants/maps.js#L12-L17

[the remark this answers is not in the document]
linxiao-li: Sounds goood, I'll give it a try!

packages/client/src/hooks/useTreasury.js
```diff
@@ -220,14 +203,21 @@ const doProposeAddVault = async (treasuryAddr, contractName) => {
...
-      arg(height, t.UInt64)
+      arg(height, t.UInt64),
     ],
     limit: SIGNED_LIMIT,
   });
 };
 
-const doProposeAddCollection = async (treasuryAddr, contractName, contractAddress) => {
-  const intent = `A.${contractAddress.replace("0x", "")}.${contractName}.Collection`;
+const doProposeAddCollection = async (
```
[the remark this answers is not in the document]
linxiao-li: I believe this is actually used for user to propose adding collection tracing back where this is being called. I believe this part got modified due to different prettify settings. I'll align mine with the codebase

packages/client/src/components/TestToolBox.js
linxiao-li: Tried to remove the curly braces, didn't work out, changing it to                           `{COIN_TYPE_TO_META[coin.coinType].displayName +": " + coin.balance}` for readability

packages/client/src/hooks/useAccount.js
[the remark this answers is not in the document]
linxiao-li: @amyliu6  @jacksonConrad Implemented the above solution, one caveat I would love to mention is this solution might break when the publicBalanceVault is configurable (for example USDC) as the publicBalanceVault might be different than what we defined in the maps.js. Assuming the configuration of USDC is not going to change frequently, this solution should work fine on testnet and mainnet

packages/client/src/constants/maps.js
```diff
@@ -14,20 +14,23 @@ export const COIN_TYPE_TO_META = {
...
     publicReceiverPath: { domain: "public", identifier: "fusdReceiver" },
+    publicBalancePath: { domain: "public", identifier: "fusdBalance" },
     storageVaultPath: { domain: "storage", identifier: "fusdVault" },
   },
   [COIN_TYPES.USDC]: {
     contractName: "FiatToken",
     displayName: "USDC",
     vaultName: "FiatToken.Vault",
     publicReceiverPath: { domain: "public", identifier: "USDCVaultReceiver" },
+    publicBalancePath: { domain: "public", identifier: "USDCVaultBalance" },
```
amyliu6, line 33: Can we add this field for the rest of the token too? Looks like `storageVaultPath` also needs to be added there. 
Until there is better ways to retrieve this metadata, we'll have to maintain this list to ensure all supported tokens has the required metadata here, or else flow will break for that token type. 
https://github.com/portto/blt-contracts/blob/e89b1af29793eb194f541a0c9a1e1adca590aaad/contracts/flow/token/BloctoToken.cdc#L189
https://github.com/StarlyIO/starly-token-contracts/blob/master/contracts/flow/contracts/StarlyToken.cdc#L107
https://flowscan.org/contract/A.d01e482eb680ec9f.REVV/overview

packages/client/src/hooks/useAccount.js
```diff
@@ -0,0 +1,61 @@
...
+      cadence: SEND_TOKENS_TO_TREASURY,
+      args: (arg, t) => [arg(treasuryAddr, t.Address), arg(amount, t.UFix64), arg(vaultPath, t.Path)],
+      limit: REGULAR_LIMIT,
+    });
+  };
+
+  //FOR TESTING UTILS ONLY, calling this method deposits tokens to treasury
+  //should only be used for testing
+  const initDepositTokensToTreasury = async (treasuryAddr) => {
+    for await (const coinType of COIN_TYPE_LIST) {
```
amyliu6, line 22: can we move this definition 
`const COIN_TYPE_LIST = [COIN_TYPES.FLOW, COIN_TYPES.FUSD, COIN_TYPES.USDC];`
to keep it inside `initDepositTokensToTreasury`?

packages/client/src/hooks/useAccount.js
```diff
@@ -0,0 +1,61 @@
...
+    const vaultPath = COIN_TYPE_TO_META[coinType].publicBalancePath;
+    const balance = await query({
+      cadence,
+      args: (arg, t) => [arg(address, t.Address), arg(vaultPath, t.Path)],
+    });
+    return balance;
+  };
+
+  const getUserBalances = async (address) => {
+    const result = COIN_TYPE_LIST.map((coinType) => ({
```
amyliu6, line 43: this function can be reused to get all supported tokens balance for Deposit scenario, can we just use the full list here: https://github.com/DapperCollectives/VESSEL/blob/763c98f1a580704e27c0509870091b3591ee2940/packages/client/src/constants/enums.js#L6-L13

packages/client/src/hooks/useAccount.js
linxiao-li: we use this list in getUserBalances as well

packages/client/src/hooks/useAccount.js
linxiao-li: BLT, STARLY and REVV are not deployed in emulators right now as far as I know, it will cause errors

amyliu6: Can we catch those errors and just logging it? We probably don't have to deploy all contracts on emulator, should be able to test it out on testnet.

KsenijaZivkovic: It will cause errors on staging as well - if the vault of that type is not added - and we try to borrow a reference before it's created. We might use instead the function for fetching treasury identifiers, and then based on the identifiers get the vault balance

linxiao-li: But this is for the address account not for the treasury

KsenijaZivkovic: oh it's just for testing 🙈 then it's not an issue at all

scripts/get_usdc_balance.cdc
```diff
@@ -0,0 +1,16 @@
+
```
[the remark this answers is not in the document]
linxiao-li: I'm leaning towards keeping it as it's a helpful script for querying the balance for development

packages/client/src/hooks/useAccount.js
linxiao-li: On staging or mainnet it is an issue as well, when user logs in with blocto or any other third party wallet, it's not guaranteed they have a FUSD, USDC, STARLY... vaults set up.  I think we should create a separate script for checking the vault existence, then query for their balance.

KsenijaZivkovic (approved):

packages/client/src/hooks/useAccount.js
amyliu6: Yea make sense! 👍 I didn’t think through on that earlier.

packages/client/src/hooks/useAccount.js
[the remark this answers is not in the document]
jacksonConrad: Awesome! I would not expect the public path to change, in general... it would be a lot of overhead for any apps using USDC (like ours) so they are definitely incentivized not to change it.

packages/client/src/hooks/useTreasury.js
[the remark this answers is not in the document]
amyliu6: Add vault/collection does not require multi-sig so it's actually not a "propose" but will take effect immediately. So make sense to remove "propose" word from function name. Created a ticket in backlog to track this refactor work: https://linear.app/dappercollectives/issue/VSL-172

amyliu6 (approved):

packages/client/src/hooks/useTreasury.js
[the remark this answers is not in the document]
linxiao-li: I see, that makes sense!

packages/client/src/hooks/useAccount.js
linxiao-li: Since https://linear.app/dappercollectives/issue/VSL-166  is effectively the same issue but only for FUSD, I'm going to change its description and tackle the above issue with this ticket

PR #86, opened 2022-08-26
switch from flowgraph => graffle and use treasury level events

mannynotfound:
- remove flowgraph and by extension `graphql-hooks` as it doesnt have event data filtering which will make rendering this component unable to scale
- switch to [GRAFFLE](https://docs.graffle.io/graffle-docs/apis/event-search-api) which is more filterable and easy to to use
- add our Company Id/Project Id for the indexer we've set up. Theses are ids are not sensitive and we want people consuming the indexer so it's included in `.env.example`
- modify `TransactionsList` components to work with GRAFFLE events - not updated to latest designs but every component should at least render without error - missing data will be addressed when our events are updated to include the metadata we need to render

image


packages/client/.env.example
```diff
@@ -1,2 +1,3 @@
-REACT_APP_FLOW_ENV=emulator
-REACT_APP_FLOWGRAPH_API_KEY=REACT_APP_FLOWGRAPH_API_KEY
+REACT_APP_FLOW_ENV=emulator
+REACT_APP_GRAFFLE_COMPANY_ID=f64c4980-ccea-45c2-9c9c-ffc160c5eca4
```
linxiao-li, line 2: Should these IDs be public?

packages/client/src/hooks/useTreasuryTransactions.js
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useState } from 'react';
+
+const COMPANY_ID = process.env.REACT_APP_GRAFFLE_COMPANY_ID;
+const PROJECT_ID = process.env.REACT_APP_GRAFFLE_PROJECT_ID;
+const NETWORK = process.env.REACT_APP_FLOW_ENV === "testnet" ? "test" : "main";
+const API_URL = `https://prod-${NETWORK}-net-dashboard-api.azurewebsites.net/api`;
+const FETCH_URL = `${API_URL}/company/${COMPANY_ID}/search?projectId=${PROJECT_ID}`;
+
+export default function useTreasuryTransactions(address) {
+  const [loading, setLoading] = useState(false);
```
linxiao-li, line 10: suggest renaming to `isLoading`

packages/client/.env.example
mannynotfound: Theses are ids are not sensitive and we want people consuming the indexer so it's included in .env.example

linxiao-li (approved):
LGTM

KsenijaZivkovic (approved):

PR #90, opened 2022-08-30
VSL-98: Added success modal for sending tokens and NFTs

KsenijaZivkovic:
## Description

In this PR, after the transaction from Things to do is confirmed, in case of sending tokens or NFTs to an address, the success modal will pop up. In case of other types of transactions, this won't happen.
Removed ExampleNFT and added ZeedzINO contract, updated tests and transactions to work with this.

Note: Contracts will need to be redeployed to testnet

## Demo / Test Result

New collection can be added like this:
Screen Shot 2022-08-30 at 12 14 23 PM
Success confirmation for sending flow tokens (only flow token icon is implemented for now, we need to add svgs for other icons)
Screen Shot 2022-09-07 at 5 29 07 PM
Success confirmation for sending NFT
Screen Shot 2022-09-07 at 5 21 54 PM

`make test`


mannynotfound:
im getting this when sending a FLOW token from treasury => wallet on testnet

image

KsenijaZivkovic:
> im getting this when sending a FLOW token from treasury => wallet on testnet
> 
> image

This happens because of this change inside contract:
`vaultID: self.recipientVault.borrow()!.getType().identifier,`
This contract needs to be redeployed to testnet in order to work the same like on localhost
Currently the vaultID is Capability<&AnyResource{A.9a0766d93b6608b7.FungibleToken.Receiver}> which doesn't give us a lot of information, and with the change to the contract this will be changed to: A.9a0766d93b6608b7.FlowToken.Vault

packages/client/src/constants/maps.js
```diff
@@ -16,6 +18,7 @@ export const COIN_TYPE_TO_META = {
     publicReceiverPath: { domain: "public", identifier: "flowTokenReceiver" },
     publicBalancePath: { domain: "public", identifier: "flowTokenBalance" },
     storageVaultPath: { domain: "storage", identifier: "flowTokenVault" },
+    icon: <FlowCoin />
```
linxiao-li, line 21: We should avoid adding component into the config file and opt in referring the icon with its path then dynamically import the icons where it's being used. I'm ok for leaving it as it is right now to get the rest of the work unblocked, I'll take a note for myself to refactor it later on [this ticket](https://linear.app/dappercollectives/issue/VSL-183)

packages/client/src/hooks/useSuccessMessage.js
```diff
@@ -0,0 +1,51 @@
+import { useEffect, useState } from "react";
+import { useModalContext } from "contexts";
+import { SuccessModal } from "modals";
+import { ASSET_TYPES } from "../constants/enums";
+
+export default function useSuccessMessage() {
```
linxiao-li, line 6: In order to avoid confusing this module with a generic success message modal, suggest renaming it to useTransactionSuccessMessage and TransactionSuccessModal,

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+
+    useEffect(() => {
+        if (type === ASSET_TYPES.NFT) {
+            const getImageURL = async () => {
+                const result = await getNFTReference(data.recipientAddr, data.nftID);
+                setImageURI(result);
+            };
+            getImageURL().catch(console.error);
+        }
+    }, [type])
```
linxiao-li, line 23: any reason to leave out the dependency of data.recipientAddr and data.nftID?

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+
+    function getNFTName(collectionId) {
```
linxiao-li, line 25: Any reason for using functions instead of just returning the value?

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+            </div>
+
+            <div className="mt-4 p-5 success-modal-background">
+                <button className="button flex-1 is-info mb-1">
+                    Sent <ArrowUp className="ml-2 has-text-white" />
+                </button>
+                <div className="pl-4">
+                    {type === ASSET_TYPES.NFT &&
+                        <>
+                            <img className="columns is-vcentered is-multiline is-mobile mr-2 mt-2 success-modal-image" src={imageURI} />
```
linxiao-li, line 60: imageURI could be undefined, suggest conditional rendering this image checking imageURI first

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+            };
+            getImageURL().catch(console.error);
+        }
+    }, [type])
+
+    function getNFTName(collectionId) {
+        return collectionId.split(".")[2];
+    }
+
+    function getTokenName(vaultId) {
```
linxiao-li, line 29: we can group this part of the logic into a getTokenMeta util function and use it like this:
`const {name, icon} = getTokenMeta(vaultId)`

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: Do we use a generic success modal?

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: It's more readable when it is defined as a function or const, what do you suggest is better for this case, to use it like this or like this
const getNFTName = (collectionId) => etc

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+            <div className="mt-4 p-5 success-modal-background">
+                <button className="button flex-1 is-info mb-1">
+                    Sent <ArrowUp className="ml-2 has-text-white" />
+                </button>
+                <div className="pl-4">
+                    {type === ASSET_TYPES.NFT &&
+                        <>
+                            <img className="columns is-vcentered is-multiline is-mobile mr-2 mt-2 success-modal-image" src={imageURI} />
+                            <span className="columns is-vcentered is-multiline is-mobile mr-2 is-size-2 is-family-monospace">
+                                #{data.nftID}
```
linxiao-li, line 62: nit, deconstructing the data object makes the code cleaner, consider `const {nftID, collectionId, amount} = data`

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Makes perfect sense, thanks!

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
+import React, { useContext, useEffect, useState } from "react";
+import { Web3Context } from "contexts/Web3";
+import { useClipboard } from "../hooks";
+import { Copy, ArrowUp, OpenNewTab, Close } from "../components/Svg";
+import { getFlowscanUrlForTransaction } from "utils";
+import { ASSET_TYPES } from "../constants/enums";
+import { COIN_TYPE_TO_META, CONTRACT_NAME_TO_COIN_TYPE } from "../constants/maps";
+
+const SuccessModal = ({ data, type, txID, onClose, safeName, safeAddress, safeOwners }) => {
```
linxiao-li, line 9: nit,
```suggestion
const SuccessModal = ({ actionData, type, txID, onClose, safeName, safeAddress, safeOwners }) => {
```

packages/client/src/pages/Safe.js
```diff
@@ -134,7 +135,11 @@ function Safe({ web3 }) {
   };
 
   const onConfirmAction = async ({ uuid }) => {
-    await executeAction(uuid).catch((error) => showErrorModal(error));
+    const events = await executeAction(uuid).catch((error) => showErrorModal(error));
+    if (events) {
+      const actions = events.filter(e => e.type.endsWith("ActionExecuted"));
```
linxiao-li, line 140: Do we have a defined list of all action types? If so we should put them in enum and use by reference

packages/client/src/hooks/useSuccessMessage.js
linxiao-li: If we have a generic error modal we should have one for success

packages/client/src/pages/Safe.js
KsenijaZivkovic: We have them defined in contracts, but there are a lot of them, and when every transaction is executed it can trigger a couple of events, we are currently interested only in two types of actions: TransferTokenToAccountActionExecuted and  TransferNFTToAccountActionCreated. In the future this might be a bigger list of actions, if we create a success screen for other transactions.

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    function getTokenIcon(vaultId) {
+        const tokenName = vaultId.split(".")[2];
+        return COIN_TYPE_TO_META[CONTRACT_NAME_TO_COIN_TYPE[tokenName]].icon;
+    }
+
+    function getSafeOwnerName(address) {
+        const owner = safeOwners.filter(owner => owner.address === address);
+        return owner && owner[0].name;
```
amyliu6, line 41: This is for rendering recipient name, right? Maybe name it `getRecipientName` as recipient might not be an owner. We also need to look for "SavedAddress/ContactList" to check if there is a name for given address. @mannynotfound  is work on Contact page. If it's difficult to coordinate in this PR we can create a ticket to pick up later.

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: Then this is ok, because error modal is triggered when the transaction fails for some reason

packages/client/src/constants/maps.js
KsenijaZivkovic: Why would placing a component into a map be an issue?

packages/client/src/hooks/useSuccessMessage.js
linxiao-li: Then we should name the error modal TransactionErrorModal, using a generic name for a component with specific use case is not scalable, imagine in the future we want to have a generic errorMessage or successMessage, we will have to do a lot of refactoring

packages/client/src/constants/maps.js
linxiao-li: 1. Config files should handle the single responsibility of giving the caller directions of where to find things or what thing the caller should find, putting the component in here would be equivalent to fetching the component and then handing it to the caller, it creates unnecessary chain of dependencies.
2. Let's say we have 7 token svgs all configured in the map, and you want to render the flow icon, you will end up loading all 7 of the svgs in the memory when you only need 1

KsenijaZivkovic: Gotcha! Thanks!

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Yeah, currently we reuse safe owners as contacts, so after Manny's work is done this can be updated

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Fixed!

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Fixed!

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Fixed!

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Updated actionData and actionType!

packages/client/src/hooks/useNFTs.js
```diff
@@ -101,6 +102,17 @@ export default function useNFTs() {
...
 
+  const getNFTReference = async (accountAddr, nftId) => {
+    const nftRef = await query({
+      cadence: GET_NFT_REF,
+      args: (arg, t) => [
+        arg(accountAddr, t.Address),
+        arg(nftId, t.UInt64),
+      ],
+    }).catch(console.error);
+    return nftRef.imageURI;
```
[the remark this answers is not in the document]
KsenijaZivkovic: Updated!

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: In case we add a generic success/error message we can rename this one

linxiao-li: How about moving the logic from this Hook inside of the SuccessModal, and really make this a generic one?

packages/client/src/modals/SuccessModal.js
linxiao-li: we can just do `NFTname = collectionId.split(".")[2]`

packages/client/src/modals/SuccessModal.js
```diff
@@ -0,0 +1,140 @@
...
+
+    function getTokenMeta(vaultId) {
+        if (vaultId) {
+            const tokenName = vaultId.split(".")[2];
+            return COIN_TYPE_TO_META[CONTRACT_NAME_TO_COIN_TYPE[tokenName]];
+        }
+    }
+
+    function getSafeOwnerName(address) {
+        const owner = safeOwners.filter(owner => owner.address === address);
```
linxiao-li, line 41: suggest using safeOwners.find() it will return the first encounter of qualified element, so we don't have to get it at index 0 later in the code

packages/client/src/modals/SuccessModal.js
linxiao-li: Is the purpose here to do nullish checking? suggest using optional chain
```suggestion
        return owner?.name;
```

packages/client/src/pages/Safe.js
jacksonConrad: My PR removes all the actions in TreasuryActions, so we will only need to ingest `ActionExecuted` event, which will have the action name + params attached.

linxiao-li: Let's create a ACTION_TYPES enum in enums.js and just put those two types there for right now
```
ACTION_TYPES = {
     TRANSFER_TOKEN_TO_ACCOUNT_ACTION_EXECUTED = "TransferTokenToAccountActionExecuted"
     TRANSFER_NFT_TO_ACCOUNT_ACTION_EXECUTED = "TransferNFTToAccountActionCreated"
}

```
and use it like this:
```suggestion
      const actions = events.filter(e => e.type == TRANSFER_TOKEN_TO_ACCOUNT_ACTION_EXECUTED);
```

KsenijaZivkovic: I'll update this once Jackson's PR gets merged

KsenijaZivkovic: Updated!

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Updated, and used contacts instead of safe owners

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Updated

packages/client/src/modals/SuccessModal.js
KsenijaZivkovic: Updated!

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: I made some modifications - renamed the hook const and the success modal to include transaction in the naming
Also, in the hook I removed useState and useEffect and tested - it looks like it works really well
Can you please take another look?

packages/client/src/modals/TransactionSuccessModal.js
```diff
@@ -0,0 +1,131 @@
...
+    const { recipient, nftId, collectionId, vaultId, tokenAmount } = actionData;
+    const { displayName, icon } = getTokenMeta(vaultId) || {};
+    const { name: imageName, imageURI } = image || {};
+
+    const NFTName = collectionId?.split(".")[2];
+
+    useEffect(() => {
+        if (actionType === ASSET_TYPES.NFT) {
+            const getImageURL = async () => {
+                const result = await getNFTReference(recipient, nftId);
```
amyliu6, line 24: `getNFTReference` is currently hardcoded with ZeedzINO, we should make it generic.

packages/client/src/hooks/useSuccessMessage.js
```diff
@@ -0,0 +1,41 @@
...
+    const { openModal, closeModal } = useModalContext();
+
+    const showSuccessTransactionModal = (action, safeName, safeAddress) => {
+        const actionData = action.data.actionView;
+        openModal(
+            <TransactionSuccessModal
+                safeName={safeName}
+                safeAddress={safeAddress}
+                actionData={actionData}
+                actionType={getActionType(actionData.type)}
```
amyliu6, line 15: since `actionType` belongs to `actionData`, do we have to pass `actionType` here?

packages/client/src/hooks/useSuccessMessage.js
```diff
@@ -0,0 +1,41 @@
...
+            {
+                headerTitle: "Success",
+            }
+        );
+    }
+
+    const getActionType = (type) => {
+        switch (type) {
+            case ACTION_TYPES.TRANSFER_NFT:
+                return ASSET_TYPES.NFT;
```
amyliu6, line 28: why do we want to map action_type to asset_type? can we just use `action_type` directly?

KsenijaZivkovic: Removed

packages/client/src/hooks/useSuccessMessage.js
amyliu6: could you help me understand why do we use `hook` instead of a normal react component for this success screen?

packages/client/src/constants/maps.js
KsenijaZivkovic: I understood that this will be addressed in [this ticket](https://linear.app/dappercollectives/issue/VSL-183)

packages/client/src/hooks/useSuccessMessage.js
```diff
@@ -0,0 +1,41 @@
+import { useModalContext } from "contexts";
+import { TransactionSuccessModal } from "modals";
+import { ASSET_TYPES, ACTION_TYPES } from "../constants/enums";
+
+export default function useSuccessMessage() {
+    const { openModal, closeModal } = useModalContext();
+
+    const showSuccessTransactionModal = (action, safeName, safeAddress) => {
```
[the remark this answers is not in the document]
KsenijaZivkovic: Fixed

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: No need, so I moved it to the modal

packages/client/src/constants/maps.js
amyliu6: ah sorry missed that.

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: Discussed with Linxiao on this topic when we implemented the error modal, based on the comment:
`Let’s encapsulate the showError and extractError into a useErrorMessage hook. Utils should not be aware of the components`
So I went on with the same approach here

packages/client/src/flow/proposeNFTTransfer.tx.js
```diff
@@ -3,7 +3,7 @@ export const PROPOSE_NFT_TRANSFER = `
 	import DAOTreasuryV4 from 0xDAOTreasuryV4
 	import NonFungibleToken from 0xNonFungibleToken
 	import MyMultiSigV4 from 0xMyMultiSigV4
-	import ExampleNFT from 0xExampleNFT
+	import ZeedzINO from 0xZeedzINO
```
amyliu6, line 6: looks like this one also needs to be generic. https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/hooks/useNFTs.js#L34.
Probably it's beyond the scope of your original ticket, is it possible to accommodate in this PR?

KsenijaZivkovic: Updated this one too, thank you! :)

packages/client/src/hooks/useNFTs.js
```diff
@@ -23,15 +24,15 @@ const doSendNFTToTreasury = async (treasuryAddr, tokenId) => {
   });
 };
 
-const doProposeNFTTransfer = async (treasuryAddr, recipient, nft) => {
-  // Example: A.f8d6e0586b0a20c7.ExampleNFT.Collection-0
+const doProposeNFTTransfer = async (contractName, contractAddress, treasuryAddr, recipient, nft) => {
+  // Example: A.f8d6e0586b0a20c7.ZeedzINO.Collection-0
```
amyliu6, line 28: looks like we can grab contract info from `nft` inside this func itself?

packages/client/src/hooks/useSuccessMessage.js
KsenijaZivkovic: I updated this to work without using hooks.

KsenijaZivkovic:
> > im getting this when sending a FLOW token from treasury => wallet on testnet
> > image
> 
> This happens because of this change inside contract: `vaultID: self.recipientVault.borrow()!.getType().identifier,` This contract needs to be redeployed to testnet in order to work the same like on localhost Currently the vaultID is Capability<&AnyResource{A.9a0766d93b6608b7.FungibleToken.Receiver}> which doesn't give us a lot of information, and with the change to the contract this will be changed to: A.9a0766d93b6608b7.FlowToken.Vault
> 
> Note: We need this change in order to know what type of token is being sent

Just to follow up on this one as well, now with the new versions of contracts -> this doesn't happen on Testnet!

amyliu6 (approved):
LGTM!

packages/client/src/modals/TransactionSuccessModal.js
```diff
@@ -0,0 +1,134 @@
...
+        if (actionType === ACTION_TYPES.TRANSFER_NFT) {
+            const getImageURL = async () => {
+                const result = await getNFTReference(NFTName, formatAddress(NFTAddress), recipient, nftId);
+                setImage(result);
+            };
+            getImageURL().catch(console.error);
+        }
+    }, [actionType, NFTName, NFTAddress, nftId, recipient, getNFTReference])
+
+    function getTokenMeta(vaultId) {
```
linxiao-li, line 34: suggest moving this to util, also create a similar function for collectionIds

packages/client/src/modals/TransactionSuccessModal.js
```diff
@@ -0,0 +1,134 @@
...
+    return (
+        <div className="p-5 has-text-black has-text-left">
+            <div className="p-5 success-modal-background">
+                <button className="button flex-1 is-info mb-1">
+                    Sent <ArrowUp className="ml-2 has-text-white" />
+                </button>
+                <div className="pl-4">
+                    {actionType === ACTION_TYPES.TRANSFER_NFT &&
+                        <>
+                            <img className="columns is-vcentered is-multiline is-mobile mr-2 mt-2 success-modal-image" src={imageURI} alt={imageName} />
```
linxiao-li, line 55: imageURI could be undefined, suggest checking it first

packages/client/src/modals/TransactionSuccessModal.js
```diff
@@ -0,0 +1,134 @@
...
+                            </span>
+                        </>
+                    }
+                </div>
+            </div>
+            <div className="mt-4 border-light-top">
+                <div className="column is-vcentered is-multiline is-mobile border-light-bottom is-flex is-full">
+                    <span className="flex-1 has-text-grey">Sent From</span>
+                    &nbsp;
+                    <div className="flex-1" style={{ position: "relative" }}>
```
linxiao-li, line 80: What is the position relative here for?

packages/client/src/modals/TransactionSuccessModal.js
```diff
@@ -0,0 +1,134 @@
...
+                            <span className="pointer" onClick={() => clipboard.copy(safeAddress)}>
+                                <Copy className="mt-1 ml-2 pointer" />
+                            </span>
+                        </div>
+                    </div>
+                </div>
+                <div className="column is-vcentered is-multiline is-mobile border-light-bottom is-flex is-full">
+                    <span className="has-text-grey flex-1">Sent To</span>
+                    &nbsp;
+                    <div className="flex-1" style={{ position: "relative" }}>
```
linxiao-li, line 97: same as above

packages/client/src/modals/TransactionSuccessModal.js
KsenijaZivkovic: Most probably won't be defined, but if it is, we will display it with an alt.
It's a good question for the team, what should we present in case something is wrong with the image -> some default image from our end or not display image at all?

packages/client/src/modals/TransactionSuccessModal.js
KsenijaZivkovic: Moved

packages/client/src/modals/TransactionSuccessModal.js
KsenijaZivkovic: There were some issues with displaying this in the same line, but now it's resolved so I removed it

packages/client/src/modals/TransactionSuccessModal.js
amyliu6: I would say not display as we don't have control on that data.

KsenijaZivkovic: Ok, added logic around that too!

linxiao-li (approved):
LGTM

PR #91, opened 2022-08-31
add ability to add/edit/remove contacts at treasury level

mannynotfound:
## Description
Reworks the contacts page to add ability to add/edit/remove contacts which are stored locally at the treasury level. Some notes:
- by default contacts should be empty until user has added some
- contacts are treated separately from safe owners in settings page - if we want to synchronize them or use the contact names we can do that work in a follow up PR
- design is not 1:1 with figma as we're still overhauling our UI kit to support new design and some components like toasts are not built yet - goal was to make the page functional while using our existing components.

## Demo / Test Result

image
image
image
image
image


KsenijaZivkovic:
When treasury is created, if we assign the names to owners - this should be added to contacts?
And when someone changes the name of the contact - it should be reflected on the owners section on safe settings.
It is possible to add multiple same contact addresses - we shouldn't allow this

amyliu6:
@KsenijaZivkovic  we'll not be syncing this contact list with owners list, it'll be treated as two separate lists. You can reference slack thread here: https://brud.slack.com/archives/C03ERT4F2MB/p1661892299277179?thread_ts=1661878670.164429&cid=C03ERT4F2MB

linxiao-li:
Are InputAddress and InputText suppose to be reusable components? If so let's create another folder under src `src/library` and put them there

packages/client/src/components/ModalHeader.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
linxiao-li, line 1: Suggest bringing this component into the Modal.js, we can render the header optionally depending on the title property passed in.

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,191 @@
-import React from "react";
-import { Person } from "./Svg";
-import { useClipboard } from "../hooks";
+import React, { useState } from "react";
```
linxiao-li, line 1: Can we break down this file into the format of one component per file? It's easier to read and maintain.

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,191 @@
...
+        <button className="button is-link mt-4" onClick={openAddModal}>
+          Add Contact <Plus className="ml-2" />
+        </button>
+      </div>
+    </section>
+  );
+}
+
+function SafeContacts({ address }) {
+  const modalContext = useModalContext();
```
linxiao-li, line 104:
```suggestion
  const {openModal, closeModal} = useModalContext();
```

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,191 @@
...
+import ModalHeader from "./ModalHeader";
+import InputAddress from "./InputAddress";
+import InputText from "./InputText";
+import { Copy, Plus } from "./Svg";
+import { useClipboard, useContacts } from "../hooks";
+import { useModalContext } from "contexts";
 
-function SafeContacts({ safeOwners = [] }) {
+function EditContactModal({ contact, onConfirm, confirmText, headerTitle }) {
+  const modalContext = useModalContext();
```
linxiao-li, line 11: modals should be opened and closed from the outside, suggest passing the closeModal function as a callback instead of bringing  in the modalContext inside of the modal itself

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,191 @@
...
+    />
+  );
+
   return (
     <>
-      <div className="column p-0 mt-5 is-flex is-full">
-        <h4 className="is-size-5">Saved addresses</h4>
+      <div className="column p-0 mt-5 is-flex is-align-items-center is-justify-content-space-between is-full">
+        <h4 className="is-size-5">Saved Addresses</h4>
+        {!isEmpty(contacts) &&
```
linxiao-li, line 142: I'm a little confused on the conditions here, would this work?
```
{isEmpty(contacts)? (<EmptyContacts/>) 
: (
      <AddContactButton/>
      <ContactList/>
)}
```

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,191 @@
...
+            <div style={{ width: minWidth }}>
+              Actions
             </div>
           </div>
-        ))}
-      </div>
+          {contacts.map((contact, index) => (
+            <div
+              className="is-flex is-align-items-center is-justify-content-space-between column is-full p-5 border-light-bottom"
+              key={index}
```
linxiao-li, line 171: We should avoid using index as key, refer to the article here :https://blog.isquaredsoftware.com/2020/05/blogged-answers-a-mostly-complete-guide-to-react-rendering-behavior/#keys-and-reconciliation

packages/client/src/components/SafeOverview.js
```diff
@@ -9,7 +9,7 @@ function SafeOverview({ allBalance }) {
     "p-5 mr-6",
   ];
 
-  const tokensWithPositiveBalance = allBalance
+  let tokensWithPositiveBalance = allBalance
```
linxiao-li, line 12: Was this erroring out?

packages/client/src/hooks/useContacts.js
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useReducer, useMemo } from "react";
+import reducer, { INITIAL_STATE } from "../reducers/contacts";
+
+const storageKey = "vessel-contacts";
```
linxiao-li, line 4: Can we move this constant to the constants/constants.js file?

packages/client/src/reducers/contacts.js
```diff
@@ -0,0 +1,40 @@
+export const INITIAL_STATE = {
+  contacts: {},
+};
+
```
linxiao-li, line 4: Could you create a document section of designed the data structure? Thanks!

packages/client/src/reducers/contacts.js
```diff
@@ -0,0 +1,40 @@
+export const INITIAL_STATE = {
+  contacts: {},
+};
+
+const reducer = (state, action) => {
+  const address = Object.keys(action.payload)[0];
+  const body = action.payload[address];
+  const newContacts = (state.contacts?.[address] ?? []).slice(0);
```
linxiao-li, line 8: Is this line to get the contact for a specific treasury? Suggest renaming to treasuryContacts

packages/client/src/reducers/contacts.js
```diff
@@ -0,0 +1,40 @@
...
+
+const reducer = (state, action) => {
+  const address = Object.keys(action.payload)[0];
+  const body = action.payload[address];
+  const newContacts = (state.contacts?.[address] ?? []).slice(0);
+
+  switch (action.type) {
+    case "SET_CONTACT": {
+      const { index, newContact } = body;
+      newContacts[index] = newContact;
```
linxiao-li, line 13: using index to locate a contact is not reliable since they change all the time, suggest using the contact address
```suggestion
      const updateIndex = treasuryContacts.findIndex(c=>c.address === newContact.address);
      treasuryContacts[updateIndex] = newContact

```

packages/client/src/components/SafeContacts.js
mannynotfound: The Add Contact in the top right only shows up after you've already added at least one contact. If you dont have any contacts it only shows up in the body underneath the "This safe doesnt have any saved addresses". 

The reason there is two checks for isEmpty(contacts) is purely the structure of the dom where the Add contact button is nested inside the always visible "Saved Addresses" masthead.

packages/client/src/components/SafeOverview.js
mannynotfound: yes this was unrelated error trying to re-assign a const

packages/client/src/components/SafeContacts.js
mannynotfound: Personally I prefer to have code logic co-located until it gets too long so i'm not bouncing between files to piece logic together. If we want to enforce this paradigm though we should add https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md to our eslint

packages/client/src/hooks/useContacts.js
mannynotfound: I would prefer to have this co-located with the reducer it belongs to, similar to how our other storage keys are set. It's only used once per hook.

packages/client/src/reducers/contacts.js
mannynotfound: I went with index manipulation vs finding specific keys because the indexes only change when you add/remove people from contacts (there is no resorting currently) so it works fine for our purposes and felt like simpler code

packages/client/src/components/SafeContacts.js
linxiao-li: I understand it's easier to work on the single file, but I think we should still choose the one component per file paradigm for the following reason:
- It's difficult to determine and enforce the threshold, everyone has different standard in terms of how long is too long
- For me it's actually easier to piece logic together with separate files since I can split panel on my editor and do cross reference, if it's one single file I'm constantly scrolling back and forth.
- When we want to add unit test for components it's easier to organize if it's one component per file.

packages/client/src/components/SafeOverview.js
linxiao-li: Thanks! I'll look into that

packages/client/src/components/SafeContacts.js
```diff
@@ -1,43 +1,203 @@
...
+
+  return (
+    <div className="py-5 has-text-black">
+      <div className="column is-flex is-flex-direction-column is-full px-5 py-0">
+        <div>
+          <p className="has-text-grey">Address<span className="has-text-red">*</span></p>
+          <InputAddress 
+            value={currentAddr} 
+            isValid={addressValid} 
+            onChange={onAddressChange} />
```
amyliu6, line 31: seems it will always allow address change? For "Edit Contact", Address field need to be read-only, only Name can be updated. https://www.figma.com/file/sKtcz6ytBmNNDNW7XU8ea5/Vessel%3A-User-Experience?node-id=330%3A188312

packages/client/src/hooks/useContacts.js
linxiao-li: From reusability perspective it does make sense to keep it close to the data, but I would argue that since this is a very high level configuration (equivalent to a db config) sprinkling them across the project makes it hard to edit and understand. Looking into the future, we probably will move into using different strategy for data storage (indexDB or a decentralized db), having configurations in one single file will help switching the strategy

packages/client/src/components/SafeContacts.js
amyliu6: yeah seems ok to allow user to edit Address too, as long as it's a valid address. @almendili what do you think?

packages/client/src/reducers/contacts.js
```diff
@@ -0,0 +1,44 @@
...
+ * where each contact object has { name: <string>, address: <string> }
+ */
+export const INITIAL_STATE = {
+  contacts: {},
+};
+
+const reducer = (state, action) => {
+  const address = Object.keys(action.payload)[0];
+  const body = action.payload[address];
+  const treasuryContacts = (state.contacts?.[address] ?? []).slice(0);
```
linxiao-li, line 12: We want to copy the state array instead of referring to it to maintain the immutibility
```suggestion
  const treasuryContacts = ([...state.contacts?.[address]] ?? []).slice(0);

```

packages/client/src/reducers/contacts.js
linxiao-li: Got it, it makes sense. I'm thinking it's probably time for us to bring in immer for handling the state update, it can help make state update simple and reliable

packages/client/src/reducers/contacts.js
mannynotfound: that's why I have `.slice(0)` which makes a copy of the array - kind of an old school method before rest spread became the norm

linxiao-li (approved):
Going to work on a coding style document pretty soon, stay tuned!

packages/client/src/components/SafeContacts.js
amyliu6: confirmed with Jose that we should allow Edit Address too. 👍 @almendili, could you help update figma design to reflect the same? Thanks!

amyliu6 (approved):

PR #92, opened 2022-08-31
VSL-78: Safe address loading fixed

KsenijaZivkovic:
## Description

This PR fixes the issue when user refreshes the page for creating safe and then try to submit the transaction.
The issue that caused this was that address isn't always set for the creator of the safe, and is sometimes undefined.

## Demo / Test Result

Bug is fixed on the UI, no errors appear when creating the safe.


jacksonConrad (approved):

packages/client/src/pages/CreateSafe.js
```diff
@@ -107,9 +107,16 @@ function CreateSafe({ web3 }) {
...
-  ]);
+  const [safeOwners, setSafeOwners] = useState();
+
+  useEffect(() => {
+    if (address) {
+      setSafeOwners([
+        { name: "", address, verified: true },
+      ])
+    }
+  }, [address]);
```
linxiao-li, line 118: If for some reason address changes, we will reset all the safeOwner data to `[{ name: "", address, verified: true }]` is this the expected behavior?

KsenijaZivkovic: Address wont change, this is just a prevention for the one point in time where address is not yet loaded and is undefined

linxiao-li: Got it, in this case we should prevent this page from rendering unless we have address ready. Using useEffect will create unexpected side effects, like you mentioned, it could change from undefined to defined, what about the situation of it changing from defined to undefined?

mannynotfound (approved):

packages/client/src/pages/CreateSafe.js
linxiao-li: If it's to deal with the catching up situation that only happens on the initial rendering, it's cheaper and more reliable to do optional rendering of the CreateSafe component since address is required to populate the page.

KsenijaZivkovic: I just pushed a different solution, with conditional rendering
In order to accomplish that I had to add a helper prop to the user 'loaded'
This helper is needed beside the address check, because when the user is not logged in - the address is undefined, then the home page won't load
The other attribute that I considered is 'loggedIn', but this one is too undefined due to async fetching
Let me know your thoughts :)

linxiao-li: I prefer the idea of using the `loggedIn`, since address is critical in creating a safe and loading a safe, preventing the home page from loading when address is absent is the exact behavior we should expect in my opinion.
Also I would suggest move this check logic from Web3 into Home.js line 36. Web3 is an HOC, having the check logic there will impact all the child components using it, some of them don't care about if user is logged in or has an address.

KsenijaZivkovic: It's not related to the Home component, it's related to the CreateSafe, so it won't work if we move the logic there.
It's not affecting the logic at all, it is just stopping it from re-rendering more times than needed - so we need to make sure the data is loaded before we render anything -> so either the user is loaded (when the user is not logged in) or the address is loaded (when the user is logged in)

linxiao-li:
@amyliu6 @KsenijaZivkovic @mannynotfound 
I've noticed that the fcl gets re-config on page load, causing the fcl to refetch user on page refreshing, causing the user data to be undefined while fcl is fetching the user data. I think this is the behavior we should avoid since config and logged in user session should remain relatively stable. I'm thinking about refactoring the web3 in the following ways:
1. Right now it's handling too many responsibilities, I'm thinking about breaking it down into UseTransaction, useTreasury, useFCLUser, useConfig
2. Reduce dependency for useFCLUser hook, have it only fetching user when necessary, have useFCLUser reads from session first before go ahead and fetching data remotely. 
Any comments, thoughts?

KsenijaZivkovic:
> @amyliu6 @KsenijaZivkovic @mannynotfound I've noticed that the fcl gets re-config on page load, causing the fcl to refetch user on page refreshing, causing the user data to be undefined while fcl is fetching the user data. I think this is the behavior we should avoid since config and logged in user session should remain relatively stable. I'm thinking about refactoring the web3 in the following ways:
> 
> 1. Right now it's handling too many responsibilities, I'm thinking about breaking it down into UseTransaction, useTreasury, useFCLUser, useConfig
> 2. Reduce dependency for useFCLUser hook, have it only fetching user when necessary, have useFCLUser reads from session first before go ahead and fetching data remotely.
>    Any comments, thoughts?

@linxiao-li I agree with the proposed refactor, it will definitely be more readable and clear who does what when it comes to responsibilities.

PR #93, opened 2022-08-31
VSL-174: Update Contract Events to be used by FE

jacksonConrad:
## Description
- Moves all events out of `MyMultiSig.cdc` and into `DAOTreasury.cdc`
- Adds `getView()` function signature to `MyMultiSig.Action` interface for passing parameterized action details to Events.
- Implements `getView()` for each Action in `TreasuryActions.cdc`
- Adds `signerApproveAction` and `signerRejectAction` calls to `DAOTreasury` as proxies for those same calls in `MyMultiSig`.  This gives us more control over the control flow (i.e. destroying the event if it has enough rejections), and better control over emitted events since we can easily include `self.uuid`.
- Include Treasury UUID + all signer responses + parameterized action details in `ActionProposed`. `ActionDestroyed`, `ActionExecuted`,`ActionApprovedBySigner`,`ActionRejectedBySigner` events.
- Updates Event Names to all be past tense

TODO: write tests to validate that the correct events are emitted with correct data

## Demo / Test Result
```
make test
```


contracts/MyMultiSig.cdc
```diff
@@ -23,20 +12,68 @@ pub contract MyMultiSigV3 {
...
+        pub var type: String
+        pub var intent: String
+        pub var proposer: Address
+        pub(set) var recipient: Address?
+        pub(set) var vaultId: String?
+        pub(set) var collectionId: String?
+        pub(set) var nftId: UInt64?
+        pub(set) var tokenAmount: UFix64?
+        pub(set) var signerAddr: Address?
+        pub(set) var newThreshold: UInt?
```
KsenijaZivkovic, line 54: What do you think about dividing this into multiple structs that inherit from ActionView?

jacksonConrad: I was thinking about doing it that way, but decided to go with one single struct because its less code

contracts/DAOTreasury.cdc
```diff
@@ -62,9 +97,18 @@ pub contract DAOTreasuryV3 {
...
+      emit ActionExecuted(
+        treasuryUUID: self.uuid,
+        actionUUID: actionUUID,
+        executor: signaturePayload.signingAddr,
+        actionView: actionView,
+        signerResponses: self.multiSignManager.getSignerResponsesForAction(actionUUID: actionUUID)
+      )
       self.multiSignManager.executeAction(actionUUID: actionUUID, {"treasury": selfRef})
-      emit ExecuteAction(actionUUID: actionUUID, proposer: signaturePayload.signingAddr)
+
```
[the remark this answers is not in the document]
jacksonConrad: i did it because the action gets destroyed so `action.signerResponses` threw an error after running `executeAction`... but i just realized i can save it to a variable before it gets destroyed, so i'll update :)

contracts/MyMultiSig.cdc
```diff
@@ -23,20 +12,68 @@ pub contract MyMultiSigV3 {
...
 
     //
-    // ------- Structs --------
+    // ------- Structs ------- 
     //
 
+    // General View Struct for Actions
+    pub struct ActionView {
+        pub var name: String
+        pub var type: String
```
linxiao-li, line 45: Can we in cadence define a enum of all action types and define the type here to be one of those enums?

contracts/TreasuryActions.cdc
```diff
@@ -69,13 +34,19 @@ pub contract TreasuryActionsV3 {
...
+    }
 
-      emit TransferTokenToAccountActionExecuted(
-        recipientAddr: self.recipientVault.borrow()!.owner!.address,
-        vaultID: self.recipientVault.getType().identifier,
-        amount: self.amount,
-        treasuryAddr: treasuryRef.owner!.address
+    pub fun getView(): MyMultiSigV3.ActionView {
+      let view: MyMultiSigV3.ActionView = TreasuryActionsV3.InitializeActionView(
+        name: "Transfer Tokens",
```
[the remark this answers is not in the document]
mannynotfound: @jacksonConrad we can keep a map on the front end of "friendly labels" as long as the type doesn't change. that way dont have to define them in contract and we have more flexibility to update the verbiage later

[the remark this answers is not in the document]
jacksonConrad: sounds good!

contracts/MyMultiSig.cdc
```diff
@@ -227,15 +238,19 @@ pub contract MyMultiSigV3 {
...
-            if self.getTotalRejectedForAction(actionUUID: actionUUID) > (self.signers.length - Int(self.threshold)) {
-                let removedAction <- self.actions.remove(key: actionUUID)
-                destroy removedAction
-                emit ActionDestroyed(uuid: actionUUID)
-            }
+        access(account) fun attemptDestroyAction(actionUUID: UInt64) {
+            pre {
+                self.canDestroyAction(actionUUID: actionUUID):
+                    "Action does not have enough rejections to destroy"
+            }
```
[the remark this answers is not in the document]
amyliu6: oh i see we have `if self.multiSignManager.canDestroyAction` condition before calling this method. Missed that earlier 😅

contracts/MyMultiSig.cdc
jacksonConrad: we can definitely do it, but i'm not sure it will actually provide many benefits.  this is because an enum ultimately maps values to a `UInt`, so we would either need to store our own map of `UInt -> String` in the script file, or in our FE javascript, so in my opinion it is more trouble than it is worth.

is there a benefit you think we are missing out on if we don't store these as Enums?

contracts/MyMultiSig.cdc
jacksonConrad: we are combining all the metadata into one struct, however I don't think there is any benefit of breaking them out into multiple structs.  the frontend will need to maintain a map of all actions and the params for each action anyway in order to display them.  so breaking out into separate structs doesn't actually provide any additional functionality or readability, unless i'm missing something.  i think the code is a bit more concise this way.

contracts/DAOTreasury.cdc
```diff
@@ -8,22 +8,38 @@ pub contract DAOTreasuryV3 {
...
-
-
-  // Interfaces + Resources
+
+  // Actions
+  pub event ActionProposed(treasuryUUID: UInt64, actionUUID: UInt64, proposer: Address, actionView: MyMultiSigV3.ActionView)
+  pub event ActionExecuted(treasuryUUID: UInt64, actionUUID: UInt64, executor: Address, actionView: MyMultiSigV3.ActionView, signerResponses: {Address: UInt})
+  pub event ActionDestroyed(treasuryUUID: UInt64, actionUUID: UInt64, signerResponses: {Address: UInt})
+  pub event ActionApprovedBySigner(treasuryUUID: UInt64, address: Address, uuid: UInt64, signerResponses: {Address: UInt})
+  pub event ActionRejectedBySigner(treasuryUUID: UInt64, address: Address, uuid: UInt64, signerResponses: {Address: UInt})
```
amyliu6, line 21: for above two events, since they already have `treasuryUUID`, can we get rid of `uuid`?

jacksonConrad: `uuid` is the action UUID! we could roll it into `actionView` although I'm not sure if we can filter events in Graffle using params nested in structs... so probably worth keeping it on the action directly.

amyliu6: it looks like they both assigned with `self.uuid` and i thought that's the treasury uuid, did i miss anything here?
https://github.com/DapperCollectives/VESSEL/blob/eb56c078bd14ba4b7d1a8e2be2b0cef84fad37bb/contracts/DAOTreasury.cdc#L72

jacksonConrad: oh nope you are totally right, sorry about that. i will change `uuid: self.uuid` -> `actionUUID: action.uuid`

amyliu6 (approved):

KsenijaZivkovic (approved):

PR #96, opened 2022-09-07
VSL-131: Added view for assets on settings page

KsenijaZivkovic:
## Description

Added view on settings page of the treasury with: 
- Buttons for adding vaults and treasuries
- Tables for presenting vaults and treasuries - from the state
- Changed vessel-assets to vessel-collections (which stores NFTs in the state)
- Added vessel-vaults which store vaults in the state 
- Remove button opens the modal for removing the vault/collection
- View on Flowscan will redirect to the contract address
- Updated modals for adding vaults and collections

## Demo / Test Result

No vaults, no collections added: 
Screen Shot 2022-09-13 at 9 39 45 AM
Modal for adding vault:
Screen Shot 2022-09-13 at 9 40 02 AM
Modal for adding vault with already existing vault:
Screen Shot 2022-09-13 at 9 40 37 AM
Modal for adding vault valid:
Screen Shot 2022-09-13 at 9 41 05 AM
Modal for adding collection:
Screen Shot 2022-09-13 at 9 41 23 AM
Modal for adding collection valid:
Screen Shot 2022-09-13 at 9 41 58 AM
Three vaults, one collection added:
Screen Shot 2022-09-13 at 9 42 23 AM
Remove vault modal:
Screen Shot 2022-09-13 at 9 42 47 AM
Remove collection modal:
Screen Shot 2022-09-13 at 9 42 59 AM


linxiao-li:
Just a heads up, all the work around button UI probably need to be refactored into using the new UI system once VSL-182 is merged, I'll work with you to make the update.

packages/client/src/components/SafeSettings/components/Assets/AddCollection.js
```diff
@@ -30,36 +31,42 @@ const AddCollection = ({ onCancel, onNext }) => {
 
   return (
     <>
-      <div className="p-5">
-        <h2 className="is-size-4 has-text-black">Add Collection</h2>
+      <div className="px-5 pt-5 columns is-vcentered is-multiline is-mobile">
+        <h2 className="is-size-4 has-text-black column">Add NFT Collection</h2>
+        <span className="pointer" onClick={onCancel}>
+          <Close className="mr-4" />
+        </span>
```
linxiao-li, line 38: The close button and header title is now handled by Modal.js as a prop passed in.

packages/client/src/components/SafeSettings/components/Assets/AddCollection.js
```diff
@@ -30,36 +31,42 @@ const AddCollection = ({ onCancel, onNext }) => {
...
       <div className="border-light-top p-5 has-text-grey">
         <div className="flex-1 is-flex is-flex-direction-column">
-          <label className="has-text-grey mb-2">Contract Name</label>
-          <input
-            className="p-4 rounded-sm border-light"
-            type="text"
-            value={contractName}
-            onChange={(e) => setContractName(e.target.value)}
-          />
+          <label className="has-text-grey mb-2">NFT Contract Address</label>
```
linxiao-li, line 42: Can we use the InputAddress.js reusable component for this part?

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
```diff
@@ -15,16 +17,19 @@ const AddVault = ({ onCancel, onNext }) => {
     displayText: type[1].displayName,
   }));
 
-  const [coinType, setCoinType] = useState(coinTypes[0].itemValue);
+  const [coinType, setCoinType] = useState("Select Token");
```
linxiao-li, line 20: Suggest 
1. renaming the state from coinType to selectedCoinType
2. initiating with a nullish value, `""` or `undefined`, 
3. prepend coinTypes with {itemValue:undefined, displayText:""}, 
4. have dropdown component accepting a defaultText property for displaying "Select Token" 
why:
"Select Token" is not one of the coin types, on validating the input, it's easier to check against a nullish value than a string

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,40 @@
...
+import { formatAddress, getFlowscanUrlForContract } from "utils";
+import { Copy } from "components/Svg";
+
+const AssetTableView = ({ assets, onRemoveClick }) => {
+    const clipboard = useClipboard();
+
+    return (
+        <div className="border-light rounded-sm py-3 table-border">
+            {assets.map(asset => {
+                const address = asset?.split(".")[1];
```
linxiao-li, line 12: Can we create a parseAsset util function? it takes the asset string as the argument and returns {address, name, ...otherStuff} that we we can just do const {address, name} = parseAsset(asset)

packages/client/src/components/SafeSettings/components/Assets/RemoveCollection.js
```diff
@@ -0,0 +1,31 @@
+import { Close } from "components/Svg";
+
+const RemoveCollection = ({ name, onCancel, onNext }) => {
+    const nextButtonClasses = ["button flex-1 p-4", "is-link"];
```
linxiao-li, line 4: Just curious on why we join the classes instead of just declare them in the className

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
```diff
@@ -0,0 +1,31 @@
+import { Close } from "components/Svg";
```
linxiao-li, line 1: Since removeVault and removeCollection are almost identical except for the message, anyway we can combine the two?

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,46 @@
...
+                        return (
+                            <div className="mx-0 p-3 columns" key={name}>
+                                <div className="column">{name}</div>
+                                <div className="column has-text-weight-bold">
+                                    {formattedAddress}
+                                    <span className="pointer" onClick={() => clipboard.copy(formattedAddress)}>
+                                        <Copy className="mt-1 ml-2 pointer" />
+                                    </span></div>
+                                <a
+                                    className="column is-flex is-justify-content-end has-text-purple button border-none"
```
linxiao-li, line 27: This a tag doesn't have any child, what are we trying to achieve for giving it flex class and justify content end?

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,46 @@
...
+                                    href={getFlowscanUrlForContract(address, name)}
+                                    target="_blank"
+                                    rel="noreferrer">
+                                    View on Flowscan
+                                </a>
+                                <button className="column is-flex is-justify-content-end has-text-purple button border-none" onClick={() => onRemoveClick(asset, name)}>
```
linxiao-li, line 33: same here

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
```diff
@@ -0,0 +1,37 @@
...
+
+const RemoveVault = ({ name, address, onCancel, onNext }) => {
+    const nextButtonClasses = ["button flex-1 p-4", "is-link"];
```
linxiao-li, line 4: Same as above, any reason to take the classes out from the dom?

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
```diff
@@ -0,0 +1,37 @@
...
+                <div className="flex-1 is-flex is-flex-direction-column">
+                    <p>
+                        This action will&nbsp;
+                        <span className="has-text-weight-bold">
+                            remove {name} ({address}) 
+                        </span>
+                        &nbsp;from the treasury. This is only possible if your balance for this token is 0.
+                    </p>
+                </div>
+                <div className="is-flex is-align-items-center mt-5">
```
linxiao-li, line 24: I suspect we don't need the align-items-center class here, could you give it a try and see if it still works?

packages/client/src/hooks/useNFTs.js
```diff
@@ -87,7 +126,7 @@ export default function useNFTs() {
       type: "SET_NFTS",
       payload: {
         [treasuryAddr]: {
-          [identifier]: tokens,
+          [identifier]: tokens || {},
```
linxiao-li, line 129: Do we know what can cause the tokens to be undefined?

packages/client/src/hooks/useNFTs.js
```diff
@@ -100,9 +139,18 @@ export default function useNFTs() {
     }).catch(console.error);
 
     const collections = (identifiers && identifiers[1]) ?? [];
+
     for (const identifier in collections) {
       await checkCollection(treasuryAddr, collections[identifier]);
+      return;
```
linxiao-li, line 145: what does the return do?

packages/client/src/hooks/useNFTs.js
```diff
@@ -100,9 +139,18 @@ export default function useNFTs() {
...
     }
+
+    dispatch({
```
linxiao-li, line 148: Why do we set the NFTs empty in the end?

packages/client/src/hooks/useTreasury.js
```diff
@@ -1,35 +1,18 @@
...
-  REMOVE_SIGNER,
-  GET_VAULT_BALANCE,
-  PROPOSE_TRANSFER,
-  ADD_VAULT,
-  ADD_COLLECTION,
-} from "../flow";
 import { COIN_TYPES } from "constants/enums";
+import { COIN_TYPE_TO_META } from "constants/maps";
+import { createSignature } from "contexts/Web3";
+import { ADD_SIGNER, EXECUTE_ACTION, GET_PROPOSED_ACTIONS, GET_SIGNERS, GET_SIGNERS_FOR_ACTION, GET_THRESHOLD, GET_TREASURY_IDENTIFIERS, GET_VAULT_BALANCE, INITIALIZE_TREASURY, PROPOSE_TRANSFER, REMOVE_SIGNER, SIGNER_APPROVE, UPDATE_SIGNER, UPDATE_THRESHOLD } from "../flow";
```
linxiao-li, line 13: Suggest prettify this section

packages/client/src/reducers/vaults.js
```diff
@@ -0,0 +1,31 @@
+export const INITIAL_STATE = {
```
linxiao-li, line 1: Suggest renaming to VAULTS_INITIAL_STATE and VaultsReducer for readability and maintainability.

packages/client/src/hooks/useTreasury.js
amyliu6: I'm trying to add Prettier to our repo to standardize formatting, PR coming soon.

amyliu6: https://github.com/DapperCollectives/VESSEL/pull/99

packages/client/src/hooks/useNFTs.js
KsenijaZivkovic: return should stop the dispatch to be executed, when we have nfts that need to be added to the state, and the dispatch which is below should reset the collection to be empty in case no nfts are found

packages/client/src/components/SafeSettings/components/Assets/AddCollection.js
KsenijaZivkovic: Moved for all the modals in this PR, there are some modals left that need to be resolved, there is a bug ticket for it https://linear.app/dappercollectives/issue/VSL-190

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
KsenijaZivkovic: Created a parseIdentifier function -> used it in all places, and removed getNFTMeta

packages/client/src/components/SafeSettings/components/Assets/RemoveCollection.js
KsenijaZivkovic: Copy pasting from other modals 🤷 😆 
Moved these classes to button className directly, and left it only on modal where the class depends on if the form is valid (Add Vault modal)

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
KsenijaZivkovic: I was also thinking about this too, but I left them separate in case in the future someone wants to differentiate them more than just the header and message.

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,46 @@
...
+    const clipboard = useClipboard();
+
+    return (
+        <div className="border-light rounded-sm py-3 table-border">
+            {
+                assets.length > 0 ?
+                    assets.map(asset => {
+                        const address = asset?.split(".")[1];
+                        const name = asset?.split(".")[2];
+                        const formattedAddress = formatAddress(address);
```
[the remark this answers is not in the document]
KsenijaZivkovic: This should be now resolved.

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
KsenijaZivkovic: By adding is-flex and is-justify-content-end this button is centered with the other text from the table and aligned to the right, same like the Remove button.

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
KsenijaZivkovic: Similar as above answer, for justifying the text to be aligned properly.

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
```diff
@@ -0,0 +1,37 @@
...
+    return (
+        <>
+            <div className="px-5 pt-5 columns is-vcentered is-multiline is-mobile">
+                <h2 className="is-size-4 has-text-black column">Remove Token Vault</h2>
+                <span className="pointer" onClick={onCancel}>
+                    <Close className="mr-4" />
+                </span>
+            </div>
+            <div className="border-light-top p-5 has-text-grey">
+                <div className="flex-1 is-flex is-flex-direction-column">
```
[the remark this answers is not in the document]
KsenijaZivkovic: Removed the div, the classes didn't do anything, same in remove collection

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
KsenijaZivkovic: Works without it, so it's removed!

packages/client/src/hooks/useNFTs.js
KsenijaZivkovic: I removed this since it's not needed, tokens will be an empty array in case there are none added

packages/client/src/hooks/useNFTs.js
KsenijaZivkovic: It's only setting them to empty in case the identifiers are empty -> the script that is being called will check the identifiers of the NFTs that belong to this collection, so we want to make sure the state is properly updated here.

KsenijaZivkovic:
> Just a heads up, all the work around button UI probably need to be refactored into using the new UI system once VSL-182 is merged, I'll work with you to make the update.

Related to this comment, I added one comment to VSL-182 to style the buttons that are already merged, just in case this PR gets stalled, in order not to block the tickets that are in prod review so they can be tested independently.
Buttons from this PR will be updated once VSL-182 is merged

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
linxiao-li: I don't think that's how they should be used. {display:flex} and {justify-content:flex-end} have impact on the children dom not the container itself, refer to this guide: https://css-tricks.com/snippets/css/a-guide-to-flexbox/
To achieve the effect you described, attach the flex classes to the parent dom (lin 19) instead

packages/client/src/components/SafeSettings/components/Assets/RemoveVault.js
linxiao-li: I see, that makes sense, but I would still say DRY principle takes the precedence in this instance. Even when we get to the situation of different content, we should be looking at passing and rendering different properties rather than create two files.

packages/client/src/hooks/useNFTs.js
KsenijaZivkovic: It was a big brain fart, I just figured too that it will work for the first index only 😆 
I rearranged the code so it is more readable and in a fail-fast manner

packages/client/src/hooks/useNFTs.js
```diff
@@ -100,6 +139,16 @@ export default function useNFTs() {
...
 
     const collections = (identifiers && identifiers[1]) ?? [];
+
+    if (collections.length === 0) {
+      dispatch({
+        type: "SET_NFTS",
+        payload: {
+          [treasuryAddr]: {},
+        },
+      });
```
linxiao-li, line 149: Forgot to return here?

KsenijaZivkovic: Yup. But even if no return is provided, the for loop will be skipped since the collection is empty

packages/client/src/hooks/useNFTs.js
```diff
@@ -100,6 +139,16 @@ export default function useNFTs() {
...
+    }
+
     for (const identifier in collections) {
       await checkCollection(treasuryAddr, collections[identifier]);
```
linxiao-li, line 153: Since each checkCollection will fire-and-forget async, having await here won't do anything, suggest remove

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,55 @@
...
+              <a
+                className="column is-flex is-justify-content-end has-text-purple button border-none"
+                href={getFlowscanUrlForContract(contractAddress, contractName)}
+                target="_blank"
+                rel="noreferrer"
+              >
+                View on Flowscan
+              </a>
+              <button
+                className="column is-flex is-justify-content-end has-text-purple button border-none"
```
linxiao-li, line 41:
```suggestion
                className="column button is-transparent"
```

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
```diff
@@ -0,0 +1,55 @@
...
+                {formattedAddress}
+                <span
+                  className="pointer"
+                  onClick={() => clipboard.copy(formattedAddress)}
+                >
+                  <Copy className="mt-1 ml-1 pointer" />
+                </span>
+              </div>
+              <a
+                className="column is-flex is-justify-content-end has-text-purple button border-none"
```
linxiao-li, line 33:
```suggestion
                className="column button is-transparent"
```

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
```diff
@@ -1,45 +1,57 @@
...
+  const vaultProps = useVaults(address);
+
+  const [coinType, setCoinType] = useState("Select Token");
+  const vaultName = COIN_TYPE_TO_META[coinType]?.vaultName;
+
+  const isFormValid =
+    vaultName &&
+    !vaultProps.vaults[address]?.find((id) => id.indexOf(vaultName) >= 0);
+
+  const isDisabled = [
```
linxiao-li, line 21:
```suggestion
  const isDisabled = isFormValid? "":"disabled"
```

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
```diff
@@ -1,45 +1,57 @@
...
-          </button>
-          <button className="button flex-1 is-primary" onClick={onNextClick}>
-            Next
-          </button>
-        </div>
+      <div className="is-flex is-align-items-center mt-6">
+        <button className="button flex-1 is-border mr-2" onClick={onCancel}>
+          Cancel
+        </button>
+        <button className={`button flex-1 is-primary ${isDisabled}`} onClick={onNextClick}>
```
linxiao-li, line 50:
```suggestion
        <button className={`button flex-1 is-primary ${isDisabled}`} disabled={!isFormValid} onClick={onNextClick}>
```

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
KsenijaZivkovic: I'll update this, tnx! But we still need the is-flex and justify-content-end in order to position this properly

packages/client/src/hooks/useTreasury.js
amyliu6: @KsenijaZivkovic  With above PR, all new changes will be auto-formatted during commit. But since this file is committed before that change, please rebase and run `npx prettier --write .` to get it formatted.

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
KsenijaZivkovic: Renamed the state, set it to undefined and added a defaultText for Dropdown component. I didn't add a new coin type with undefined value to coin types because that would make it possible to reselect and would appear in a dropdown. Please take a look and tell me what do you think about the new approach :)

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
KsenijaZivkovic: Is it okay if we go with the **pointer-events: none** for every button that has a class disabled?
I added this to App.sass, since I noticed that we have this class appended in a couple of places, but the button wasn't actually disabled -> so this will work for all of them

amyliu6 (approved):

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
linxiao-li: Love the new approach!

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
linxiao-li: I think we should use the "disabled" class purely for styling purpose and use semantic HTML syntax. Bulma also deprecated the "is-disabled" class in favor of html disabled tag (https://bulma.io/documentation/elements/button/)

packages/client/src/reducers/vaults.js
KsenijaZivkovic: I skipped this because it's unified across the whole app, I don't feel like it's affecting readability and maintainability

packages/client/src/components/SafeSettings/components/Assets/AssetTableView.js
KsenijaZivkovic: Instead of using is-flex, I removed the is-flex, justify content and column classes, the 'column' class was messing with the position of the button. It's properly aligned now.

packages/client/src/components/SafeSettings/components/Assets/AddVault.js
KsenijaZivkovic: Updated!

packages/client/src/components/SafeSettings/components/Assets/index.js
```diff
@@ -1,56 +1,173 @@
...
+          className="button is-secondary is-small with-icon"
+          onClick={openAddVaultModal}
+        >
+          Add Token Vault
+          <Plus style={{ position: "relative", left: 5 }} />
+        </button>
+      </div>
+      <div className="column p-0 mt-4 is-flex is-flex-direction-column is-full rounded-sm border-light has-shadow">
+        <AssetTableView
+          assets={vaults && vaults[address] && Object.values(vaults[address])}
```
linxiao-li, line 148: I'm confused on this line, looking at AssetTableView, we are expecting a array, this looks like a boolean to me

packages/client/src/reducers/vaults.js
linxiao-li: I'm gonna be nit picking here and insist on renaming
Readability is when a person with no former context of our project jumping in and being able to perform static analysis of the code. Imagine yourself as that person and read where the reducer is being called, for example useNFT.js, instead of just telling you we are using the nftReducer, the code directs you to the import statement   , only then you can figure out what's going on. 
I would say there's always room for improvement and optimization even though a pattern is wildly used.

KsenijaZivkovic: Renamed! We will then need to rename the others too, can you please create a ticket for this?

packages/client/src/components/SafeSettings/components/Assets/index.js
KsenijaZivkovic: Changed the condition to
`vaults[address] && Object.values(vaults[address])`
vaults[address] will be either undefined, empty array or array with some values -> and when it comes to rendering the AssetTableView it's useful to let the component handle if the vaults are empty or the rows should be rendered.

packages/client/src/reducers/vaults.js
linxiao-li: created and assigned to myself https://linear.app/dappercollectives/issue/VSL-204

linxiao-li (approved):
LGTM! Great job!

almendili:
Front-end feedback:

- Remove shadow from token vaults and nft collection table.
- Make sure dropdown and text field has the same height.
- Add 1px border above buttons similar to the border under the modal header.

PR #99, opened 2022-09-09
Added pre-commit hook to format files

amyliu6:
## Description

In this PR,
1. Ran `npx prettier --write .` formatted existing files.
2. Added pre-commit hook to format files. In future it will auto format files during commit.


linxiao-li:
One thing caused a lot of formatting difference is the print-width, can we standardize it to 100 or 80?

amyliu6:
> One thing caused a lot of formatting difference for me is the print-width, can we standardize it to 100 or 80?

@linxiao-li , yes, default is 80. https://prettier.io/docs/en/options.html

.prettierrc.json
```diff
@@ -0,0 +1 @@
+{}
```
KsenijaZivkovic, line 1: Do we need to add here some basic configuration like:
```
{
  "trailingComma": "es5",
  "tabWidth": 4,
  "semi": false,
  "singleQuote": true
}
```
?

linxiao-li (approved):

.prettierrc.json
amyliu6: I left it empty because the default value seems already pretty aligned with our current convention. I ran `yarn prettier --write "packages/client/src/**/*.js"` and created second commit with updated files.

KsenijaZivkovic: So how does it work, do we need to run this command before commit or it's formatting the file on save?
Just thinking from the perspective if we need to add this command to scripts in package.json

KsenijaZivkovic (approved):

PR #102, opened 2022-09-12
vsl 183: lazy loading icons and organize reusable components

linxiao-li:
## Description
1. Implemented dynamic import and lazy loading for svgs, 
- to avoid importing all the svg files into the component when we only need 1 of them, especially for token icons
- easier to manage and organize

2.Moved Dropdown, ProgressBar and InputAddress into `library/components` and updated the references
-  ticket for applying the reusable component InputAddress https://linear.app/dappercollectives/issue/VSL-194

** note: I've also created a set of tickets for implementing storybook as part of the effort in building and maintaining our front-end library https://linear.app/dappercollectives/issue/VSL-195
## Demo / Test Result

Before lazy loading:
Webpack load the entire svg file into the bundle:
Screen Shot 2022-09-12 at 11 15 21 AM

After lazy loading:
Webpack load the loading helper Svg/index.js into the bundle
Screen Shot 2022-09-12 at 11 16 40 AM
When we need a svg, we make network requests for the specific file
Screen Shot 2022-09-12 at 11 17 10 AM


packages/client/src/components/SafeContacts.js
```diff
@@ -1,8 +1,7 @@
 import React, { useState } from "react";
 import { isEmpty } from "lodash";
-import InputAddress from "./InputAddress";
-import InputText from "./InputText";
```
KsenijaZivkovic, line 4: Why did we delete this?

linxiao-li: InputAddress is moved to library/InputAddress, 
InputText is only used on this page and it performs exactly like a normal text input, no reason to create a separate file for it. On the other hand, if we want to normalize the input behaviors in the future, we can create inputText reusable component and apply it across the board

KsenijaZivkovic: It would be useful to create a new component or modify the one that is deleted so it can also accept things like:
- label
- flag - if the field is required (to show the *)
- placeholder
- onChange
That way we will unify the look and have less code written.

linxiao-li: 100% agree. Would you mind creating a ticket for this?

KsenijaZivkovic: https://linear.app/dappercollectives/issue/VSL-202

packages/client/src/components/SafeContacts.js
```diff
@@ -183,11 +190,16 @@ function SafeContacts({ address }) {
       <div className="column p-0 mt-5 is-flex is-align-items-center is-justify-content-space-between is-full">
         <h4 className="is-size-5">Saved Addresses</h4>
         {!isEmpty(contacts) && (
+<<<<<<< HEAD
```
KsenijaZivkovic, line 193: Conflict

linxiao-li: Updated! Thanks a lot!

packages/client/src/components/SafeSettings/components/Assets/AddCollection.js
```diff
@@ -53,7 +53,7 @@ const AddCollection = ({ onCancel, onNext }) => {
               />
               {addressValid && (
                 <div style={{ position: "absolute", right: 17, top: 14 }}>
-                  <Check />
+                  <Svg nam="Check" />
```
KsenijaZivkovic, line 56: typo

KsenijaZivkovic:
There are some compiling errors:
- SafeSettings/components/Owners 
- TransactionSuccesModal
And one warning related to useEffect in Svg/index.js

KsenijaZivkovic (approved):

packages/client/src/library/Svg/index.js
```diff
@@ -0,0 +1,19 @@
...
+      try {
+        const svgComponent = await import(`./${name}`);
+        setSvgComponent(svgComponent.default(props));
+      } catch (e) {
+        console.log(e);
+      }
+    })();
+    //eslint-disable-next-line
+  }, [name]);
+  if (!svgComponent) return null;
```
mannynotfound, line 16: `svgComponent` should already be null if component fails to load so this line shouldn't be necessary

packages/client/src/library/Svg/index.js
```diff
@@ -0,0 +1,19 @@
...
+  useEffect(() => {
+    (async () => {
+      try {
+        const svgComponent = await import(`./${name}`);
+        setSvgComponent(svgComponent.default(props));
+      } catch (e) {
+        console.log(e);
+      }
+    })();
+    //eslint-disable-next-line
```
mannynotfound, line 14:
```suggestion
    // eslint-disable react-hooks/exhaustive-deps
```

mannynotfound (approved):
This is really cool!

I'm working on something similar for CAST that also adds `storybook` so would be good to re-use this code and maybe even share the same codebase at some point, otherwise this LGTM!

https://github.com/DapperCollectives/CAST/pull/449

packages/client/src/library/Svg/index.js
```diff
@@ -0,0 +1,18 @@
...
+const Svg = ({ name, ...props }) => {
+  const [svgComponent, setSvgComponent] = useState(null);
+  useEffect(() => {
+    (async () => {
+      try {
+        const svgComponent = await import(`./${name}`);
+        setSvgComponent(svgComponent.default(props));
+      } catch (e) {
+        console.log(e);
+      }
```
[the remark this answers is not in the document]
linxiao-li: I don't think we should use mixed syntax, it causes confusion and makes the code harder to maintain, this answer from stack overflow explains it well: https://stackoverflow.com/a/54387912.

linxiao-li:
@mannynotfound Thanks a lot! Does the design of CAST look similar to VESSEL? If so we should promote @cast/shared-components to @dapperCollectives/shared-components, it would save VESSEL a ton of development work reinventing the wheels!

PR #103, opened 2022-09-12
VSL-163: AddSignerUpdateThreshold and RemoveSignerUpdateThreshold actions

jacksonConrad:
## Description

- Adds `AddSignerUpdateThreshold` and `RemoveSignerUpdateThreshold` actions to `treasury_actions.cdc`
- Replace `TestAddSignerAction` test with `TestAddSignerUpdateThresholdAction`
- Replace `TestRemoveSignerAction` test with `TestRemoveSignerUpdateThresholdAction`

Now it is possible to add/remove a signer to/from a treasury AND update the threshold in the same action.

Once FE is integrated we can remove `AddSigner` and `RemoveSigner` actions.

## Demo / Test Result

`make test`

### DIFF
--- a/contracts/TreasuryActions.cdc
+++ b/contracts/TreasuryActions.cdc
@@ -293,4 +293,83 @@ pub contract TreasuryActionsV4 {
       self.intent = "Update the threshold of signers to ".concat(threshold.toString()).concat(".")
     }
   }
+
+  pub struct AddSignerUpdateThreshold: MyMultiSigV4.Action {
+    pub let threshold: UInt
+    pub let signer: Address
+    pub let intent: String
+    pub let proposer: Address
+
+    access(account) fun execute(_ params: {String: AnyStruct}) {
+      let treasuryRef: &DAOTreasuryV4.Treasury = params["treasury"]! as! &DAOTreasuryV4.Treasury
+      let manager = treasuryRef.borrowManager()
+
+      // Add Signer
+      manager.addSigner(signer: self.signer)
+      log("\n\nEXECUTINGGGGG!!!!!!!!!\n\n")
+      // Update Threshold
+      manager.updateThreshold(newThreshold: self.threshold)
+    }
+
+    pub fun getView(): MyMultiSigV4.ActionView {
+      let view: MyMultiSigV4.ActionView = TreasuryActionsV4.InitializeActionView(
+        type: "AddSignerUpdateThreshold",
+        intent: self.intent,
+        proposer: self.proposer
+      )
+      view.newThreshold = self.threshold
+      view.signerAddr = self.signer
+      return view
+    }
+
+    init(signer: Address, threshold: UInt, proposer: Address) {
+      self.signer = signer
+      self.threshold = threshold
+      self.proposer = proposer
+      self.intent = "Add signer "
+        .concat(signer.toString())
+        .concat(". Update the threshold of signers to ")
+        .concat(threshold.toString()).concat(".")
+    }
+  }
+
+  pub struct RemoveSignerUpdateThreshold: MyMultiSigV4.Action {
+    pub let threshold: UInt
+    pub let signer: Address
+    pub let intent: String
+    pub let proposer: Address
+
+    access(account) fun execute(_ params: {String: AnyStruct}) {
+      let treasuryRef: &DAOTreasuryV4.Treasury = params["treasury"]! as! &DAOTreasuryV4.Treasury
+      let manager = treasuryRef.borrowManager()
+
+      // Remove Signer
+      manager.removeSigner(signer: self.signer)
+      // Update Threshold
+      manager.updateThreshold(newThreshold: self.threshold)
+    }
+
+    pub fun getView(): MyMultiSigV4.ActionView {
+      let view: MyMultiSigV4.ActionView = TreasuryActionsV4.InitializeActionView(
+        type: "RemoveSignerUpdateThreshold",
+        intent: self.intent,
+        proposer: self.proposer
+      )
+      view.newThreshold = self.threshold
+      view.signerAddr = self.signer
+      return view
+    }
+
+    init(signer: Address, threshold: UInt, proposer: Address) {
+      self.signer = signer
+      self.threshold = threshold
+      self.proposer = proposer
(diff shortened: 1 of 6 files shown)


contracts/TreasuryActions.cdc
```diff
@@ -293,4 +293,82 @@ pub contract TreasuryActionsV4 {
...
+    pub let proposer: Address
+
+    access(account) fun execute(_ params: {String: AnyStruct}) {
+      let treasuryRef: &DAOTreasuryV4.Treasury = params["treasury"]! as! &DAOTreasuryV4.Treasury
+      let manager = treasuryRef.borrowManager()
+
+      // Add Signer
+      manager.addSigner(signer: self.signer)
+      // Update Threshold
+      manager.updateThreshold(newThreshold: self.threshold)
```
linxiao-li, line 310: Looks like the actions are take place async, could be problematic, consider this scenario:
- Treasury starts with one signer
-  Signer proposes to add another signer and increase the threshold to 2
-  The action of adding the signer succeeds
-  The action of updating the threshold fails since the post condition demands signer.length < threshold
-  Error message of failed updating threshold is surfaced to FE when the action is partially failed. 
-  User refreshes the page and see signer length is increased but not the threshold.

contracts/TreasuryActions.cdc
```diff
@@ -293,4 +293,82 @@ pub contract TreasuryActionsV4 {
...
+    pub let signer: Address
+    pub let intent: String
+    pub let proposer: Address
+
+    access(account) fun execute(_ params: {String: AnyStruct}) {
+      let treasuryRef: &DAOTreasuryV4.Treasury = params["treasury"]! as! &DAOTreasuryV4.Treasury
+      let manager = treasuryRef.borrowManager()
+
+      // Add Signer
+      manager.addSigner(signer: self.signer)
```
KsenijaZivkovic, line 308: What do you say about creating a one manager function to do both?

main_test.go
```diff
@@ -597,7 +597,7 @@ func TestSignerRevokeApproval(t *testing.T) {
 	})
 }
 
-func TestAddSignerAction(t *testing.T) {
+func TestAddSignerUpdateThresholdAction(t *testing.T) {
```
KsenijaZivkovic, line 600: Do we still have the possibility to call them separately? If we do, then we need both test cases. 
And some negative scenarios too

contracts/TreasuryActions.cdc
amyliu6: @linxiao-li  the post condition for `updateThreshold` is checking for `signer.length >= threshold`, which looks correct and i don't see issues there. Did i miss anything here?

linxiao-li: @amyliu6 base on what I read about the post conditions, it seems like the scope is limited to the function, in this case, line 310 only. At that point, line 308 executed successfully, line 310 fails, the whole transaction is half success half fail, but was treated as 100 percent fail because of the post condition on 310.

jacksonConrad: @linxiao-li the post-condition of `updateThreshold` is asserting that `self.signers.length >= self.threshold` is `true`.  In the scenario you have described, the second signer would be added first, so `self.signers.length` would equal 2.  Then, the threshold is updated to 2, and the post-condition will hold true.

So I think for addSigner/updateThreshold we are fine because you can always add a signer no matter the threshold.

But for removeSigner/updateThreshold, I believe there is an issue.... we should updateThreshold FIRST, then remove the signer after.

If we stick to the correct order of add signer -> update threshold, update threshold -> remove signer, I think we should be fine.  Let me know if you think there is an issue I'm not seeing

contracts/TreasuryActions.cdc
jacksonConrad: I was thinking about this because we could consolidate the 2 actions AddSignerUpdateThreshold and RemoveSignerUpdateThreshold into one.

It would DRY the code a little bit, but it would be a small optimization, and I like the clarity of having two explicitly separate actions.

contracts/TreasuryActions.cdc
linxiao-li: Got it! I misunderstood the syntax on the post condition, it is saying signer count has to be equal or greater than threshold, otherwise throw an error. That makes a lot more sense! But the issue still exists for the add signer flow, what if the param is add one more signer and increase the threshold to 3? (I know on FE there's logic to forbid such action, but don't trust us :)))) Can we add a check function before? Something like
```
fun canUpdate(newSignerCount:UInt, newThreshold:Uint): Bool {
    return newSignerCount >= newThreshold;
}
if canUpdate {
    manager.updateSigner(signer: self.signer);
    manager.updateThreshold(newThreshold: self.threshold);
}
```

jacksonConrad: Yup exactly! Thanks for pointing it out, though because it made me catch an issue w/ `RemoveSignerUpdateThreshold` (just pushed a fix!) :)

I don't think the issue exists for add signer flow, because the `updateThreshold` post-condition will always cause the entire transaction to fail.  Even if `manager.addSigner` runs successfully, the tx will fail on `manager.updateThreshold` because `self.signers.length >= self.threshold` will not hold true.

Are you saying you want to prevent a user from even proposing such an action? I'm pretty confident that the post-conditions will prevent this scenario from ever occurring (i.e. such an action will not be able to be executed if it violates this rule)

linxiao-li: > post-condition will always cause the entire transaction to fail. Even if manager.addSigner runs successfully, the tx will fail on manager.updateThreshold because self.signers.length >= self.threshold will not hold true.

Would it roll back the signer amount as well?

linxiao-li: Just did some research, looks like when there's an error for the transaction, the incurred changes get reverted. @jacksonConrad Is that correct? If that's the case I'm all good. Sorry for the confusion!

jacksonConrad: Yup, that is correct! No worries!

linxiao-li (approved):

KsenijaZivkovic (approved):

PR #104, opened 2022-09-13
VSL-135: Add Tokens Page

linxiao-li:
## Description
- Adds the tokens page
- Apply the h2s of design specs
- Create and Implement EmptyTableWithCTA reusable component
- Add auto scroll feature to safe settings page
**this pr doesn't include the functionality for depositing a token**

## Demo / Test Result

- user flow for clicking the manage token vault button and send a token from the tokens page
 
https://user-images.githubusercontent.com/7423845/190012049-fde41bd0-f530-4d20-9462-fb8d30d85f42.mov

- user flow for clicking the manage token vault button on empty vault, auto scroll to tokens section

https://user-images.githubusercontent.com/7423845/190235123-038fa811-363a-456d-92f5-b713622f08f7.mov


packages/client/src/components/SafeAssets.js
```diff
@@ -67,7 +67,6 @@ function SafeAssets({ web3, name, address }) {
             onClick={() =>
               modalContext.openModal(
                 <SendTokens
-                  name={name}
```
linxiao-li, line 70: removed the name property since that's not being used in the new design of sendToken flow

packages/client/src/components/SafeSettings/components/Assets/index.js
```diff
@@ -35,7 +35,7 @@ const Assets = ({ treasury, proposeAddVault, proposeAddCollection }) => {
     );
   };
   return (
-    <div className="p-0 mt-5">
+    <div id="assets" className="p-0 mt-5">
```
linxiao-li, line 38: We are giving the dom an id so that we can navigate directly to this component

packages/client/src/components/SafeTokens/index.js
```diff
@@ -0,0 +1,54 @@
...
+        address={treasuryAddress}
+        initialState={{
+          assetType: ASSET_TYPES.TOKEN,
+          coinType,
+        }}
+      />
+    );
+  };
+  const handleManageTokenVaults = () => {
+    history.push(`/safe/${treasuryAddress}/settings#assets`);
```
linxiao-li, line 32: Will need to change the hash location once [vsl-131](https://github.com/DapperCollectives/VESSEL/pull/96) is merged since this pr breaks the asset into two parts

packages/client/src/components/SafeTokens/EmptyVault.js
```diff
@@ -0,0 +1,12 @@
+const EmptyVault = () => (
+  <div
+    style={{
+      height: "calc(100vh - 350px)",
+    }}
+    className="is-flex is-justify-content-center is-align-items-center is-flex-direction-column"
+  >
+    <h2>You don't have any Vault</h2>
+    <p className="has-text-grey">Create a new transaction above</p>
```
amyliu6, line 9: the UI seems not as per this design: https://www.figma.com/file/sKtcz6ytBmNNDNW7XU8ea5/Vessel%3A-User-Experience?node-id=344%3A203543, could you help cross check?

linxiao-li: Thanks for the check, I for sure overlooked. Will update!

amyliu6:
@linxiao-li  thanks for updating the PR! In above demo video, it looks like user has Vaults but it's not displayed on Tokens page? The designed behavior is to display Vaults on Tokens page even if they are empty. https://www.figma.com/file/sKtcz6ytBmNNDNW7XU8ea5/Vessel%3A-User-Experience?node-id=4%3A6925

linxiao-li:
@amyliu6 Oh I programmatically set the vaults=[] to demo the empty page. It should behave just like the design

linxiao-li:
@KsenijaZivkovic Had to break down the asset component into two in order to pass the ref

amyliu6 (approved):

KsenijaZivkovic:
When user has three vaults after creating the safe - the tokens page should contain those three tokens.
When user adds or removes a vault - this should be visible on the token page immediately.
If user deletes all vaults from the settings page, the tokens page should be empty.
Is this scoped for this feature or we have a ticket for this?

packages/client/src/components/SafeSettings/index.js
```diff
@@ -24,22 +25,35 @@ const SafeSettings = () => {
     removeVault,
     removeCollection,
   } = web3;
-
+  useEffect(() => {
+    (() => {
```
KsenijaZivkovic, line 29: Looks like it would work without the extra brackets:
```javascript
useEffect(() => {
    if (targetElementRef.current) {
      targetElementRef.current.scrollIntoView({ behavior: "smooth" });
    }
 //eslint-disable-next-line
 }, []);
  ```

linxiao-li: You are right. I got used to wrap async calls to IIFE, it's not needed here 👍

linxiao-li:
@KsenijaZivkovic That's a great point! Just realized with the work you did with the vaults, this page shouldn't be dependent on the balances in treasury (which by default grabs all coin types in the map). Will make the change!

linxiao-li:
@KsenijaZivkovic updated, demo here:

https://user-images.githubusercontent.com/7423845/190515667-7a84c61c-3db5-4d08-a21a-eb7830f1950f.mov

packages/client/src/components/SafeSettings/index.js
KsenijaZivkovic: Please check if you pushed this change, it's not reflected on the PR yet

KsenijaZivkovic (approved):

almendili:
There's a few discrepancies between front-end and design for the modals:
- Overall: add 1px border above the buttons, similar to the border below the modal header.
- Send modal. The order of things are little off here. Link to figma: https://www.figma.com/file/sKtcz6ytBmNNDNW7XU8ea5/Vessel%3A-User-Experience?node-id=412%3A61216

amyliu6:
> There's a few discrepancies between front-end and design for the modals:
> 
> * Overall: add 1px border above the buttons, similar to the border below the modal header.
> * Send modal. The order of things are little off here. Link to figma: https://www.figma.com/file/sKtcz6ytBmNNDNW7XU8ea5/Vessel%3A-User-Experience?node-id=412%3A61216

Created ticket to track this work: https://linear.app/dappercollectives/issue/VSL-219

PR #107, opened 2022-09-15
VSL-204: fix flowfee estimation bug and rename reducer

linxiao-li:
## Description
1. rename the reducers for readability
2. fix the bug related to get flow fee estimation


flow.json
```diff
@@ -20,6 +20,13 @@
         "testnet": "0x7e60df042a9c0868"
       }
     },
+    "FlowFees": {
```
linxiao-li, line 23: Some how this part got removed from the flow.json, adding this back

linxiao-li:
Removed reducers/index.js since it's not being used

amyliu6 (approved):

KsenijaZivkovic (approved):

linxiao-li:
@KsenijaZivkovic Looks like we both need some coffee...had to revert the brackets otherwise the function won't get called ☕

PR #108, opened 2022-09-20
VSL-120: Approve/Reject Proposed Action

KsenijaZivkovic:
## Description

- Added script for fetching action view = it contains all useful data about the action, like vaultId, collectionId, amounts etc
- Expanded action view with a timestamp
- Opens modal when user clicks on Sign - from things to do
- This modal contains Approve and Reject buttons
- In case of reject, if it's not possible to met the threshold - action will automatically be deleted - NOTE: this is not a bug, this is a feature!

### BE
`make test`
### FE

https://user-images.githubusercontent.com/20346621/191655144-66b5c9fd-6b45-444f-8005-aec7f9d9df53.mov

https://user-images.githubusercontent.com/20346621/191656098-11f7f9f6-dfe8-4bfd-86bb-d3b8cb328286.mov


packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <div className="p-5 has-text-grey">
+      {getActionView()}
+      <Confirmations confirmations={confirmations} safeData={safeData} />
+      <div className="column border-light-bottom pb-5">
+        To complete this action, you will have to confirm it with your connected
+        wallet on the next step.
+      </div>
+      <div className="is-flex is-align-items-center mt-5">
```
linxiao-li, line 51: is-align-item-center can be removed

packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -0,0 +1,62 @@
...
+
+const ActionRequired = ({
+  safeData,
+  actionView,
+  confirmations,
+  onReject,
+  onApprove,
+}) => {
+  const actionType = actionView.type;
+  console.log(actionView);
```
linxiao-li, line 16: console log can be removed

packages/client/src/components/Actions/ActionsList.js
```diff
@@ -1,9 +1,50 @@
...
+  const onRejectAction = async (action) => {
+    try {
+      await onReject(action);
+      closeModal();
+    } catch (error) {
+      showErrorModal(error);
+    }
+  };
+
+  const openSignRejectModal = async (action) => {
```
linxiao-li, line 34: Looks like the modal opened up allows user to sign to approve OR reject, suggest rename to `openApproveOrRejectModal`

packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -0,0 +1,62 @@
...
+import Confirmations from "./Confirmations";
+
+const ActionRequired = ({
+  safeData,
+  actionView,
+  confirmations,
+  onReject,
+  onApprove,
+}) => {
+  const actionType = actionView.type;
```
linxiao-li, line 15: Prefer use destructuring
```suggestion
  const {type: actionType } = actionView
```

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,46 @@
+import Svg from "library/Svg";
+import { useContacts } from "hooks";
+import { getSafeContactName, shortenAddr, shortenName } from "utils";
+
+const Confirmations = ({ confirmations, safeData }) => {
+  const { address: safeAddress } = safeData;
+
+  const { contacts } = useContacts(safeAddress);
```
linxiao-li, line 8: As far as I know contacts are standalone state, it doesn't contain the data of safeOwners. When I'm signing an action of sending a token, I'm expecting to see my name on the confirmation block but I'm only seeing the address. Is this the expected behavior?

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,46 @@
...
+    )?.length;
+
+    return `${numberApproved} out of ${Object.keys(confirmations).length}`;
+  };
+
+  const getConfirmationList = () => {
+    return Object.keys(confirmations).map((key) => {
+      const name = getSafeContactName(contacts, key);
+      return (
+        <div className="confirmation">
```
linxiao-li, line 22: missing key property

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,46 @@
...
+
+const Confirmations = ({ confirmations, safeData }) => {
+  const { address: safeAddress } = safeData;
+
+  const { contacts } = useContacts(safeAddress);
+
+  const getNumberOfConfirmations = () => {
+    const numberApproved = Object.keys(confirmations).filter(
+      (key) => confirmations[key] === "approved"
+    )?.length;
```
linxiao-li, line 13: Probably don't need the optional chain here, if confirmations is an empty object, Object.keys(confirmations) will return an empty array, filtering it returns an empty array as well
If we want to check against nullish value of confirmations, we can do
`Object.keys(confirmations ?? [])`

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,46 @@
...
+    return `${numberApproved} out of ${Object.keys(confirmations).length}`;
+  };
+
+  const getConfirmationList = () => {
+    return Object.keys(confirmations).map((key) => {
+      const name = getSafeContactName(contacts, key);
+      return (
+        <div className="confirmation">
+          <Svg name="Status" className={`${confirmations[key]}`} />
+          {name?.length > 20 ? shortenName(name) : name}
```
linxiao-li, line 24: nit, suggest adding a couple more references for readability
```suggestion
         const displayAddress = name ? `·${shortenAddr(key)}` : key;
      const displayName = name?.length > 20 ? shortenName(name) : name;
      return (
        <div className="confirmation">
          <Svg name="Status" className={`${confirmations[key]}`} />
          {name ? `${displayName}${displayAddress}` : `${displayAddress}`}
        </div>
      );
```

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,46 @@
...
+      (key) => confirmations[key] === "approved"
+    )?.length;
+
+    return `${numberApproved} out of ${Object.keys(confirmations).length}`;
+  };
+
+  const getConfirmationList = () => {
+    return Object.keys(confirmations).map((key) => {
+      const name = getSafeContactName(contacts, key);
+      return (
```
linxiao-li, line 21: nit, suggest adding a couple more references for readability
```suggestion
         const displayAddress = name ? `·${shortenAddr(key)}` : key;
      const displayName = name?.length > 20 ? shortenName(name) : name;
      return (
        <div className="confirmation">
          <Svg name="Status" className={`${confirmations[key]}`} />
          {name ? `${displayName}${displayAddress}` : `${displayAddress}`}
        </div>
      );
```

packages/client/src/App.sass
```diff
@@ -432,3 +435,22 @@ hr
...
+  font-size: 0.875rem
+  box-shadow: none
+  border: 1px solid $black
+  border-radius: 8px
+  margin-right: 8px
+  svg
+    margin-right: 12px
+.confirmations
+  overflow: auto
+svg
```
linxiao-li, line 450: Let's have this part of svg styling nested under the confirmation class so it doesn't pollute the global style

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
+import { useClipboard, useContacts } from "hooks";
+import Svg from "library/Svg";
+import { getSafeContactName, getTokenMeta } from "utils";
+import ProposedDateView from "./ProposedDateView";
+
+const TokenView = ({ actionView, safeData }) => {
+  const { recipient, vaultId, tokenAmount, timestamp } = actionView;
+  const { name: safeName, address: safeAddress } = safeData;
+
+  const { displayName, tokenType } = getTokenMeta(vaultId) || {};
```
linxiao-li, line 10: What's the purpose of the empty object in the end?

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
...
+
+  const clipboard = useClipboard();
+  const { contacts } = useContacts(safeAddress);
+
+  return (
+    <>
+      <div className="pl-4 mb-5">
+        <span className="columns">Amount</span>
+        <span className="columns is-size-2 is-family-monospace has-text-black">
+          {Number(tokenAmount)}
```
linxiao-li, line 20:
```suggestion
          {Number(tokenAmount).toLocaleString()}
```
This will make sure the number gets delaminated with comma

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
...
+
+  const { displayName, tokenType } = getTokenMeta(vaultId) || {};
+
+  const clipboard = useClipboard();
+  const { contacts } = useContacts(safeAddress);
+
+  return (
+    <>
+      <div className="pl-4 mb-5">
+        <span className="columns">Amount</span>
```
linxiao-li, line 18: This span has no children element, what are we trying to achieve with the columns class?

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
...
+        <span className="columns is-size-2 is-family-monospace has-text-black">
```
linxiao-li, line 19: Same here

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
...
+          {Number(tokenAmount)}
+        </span>
+        <span className="columns is-size-6 has-text-weight-bold has-text-black">
+          <Svg name={tokenType} /> &nbsp; {displayName}
+        </span>
+      </div>
+      <div className="mt-4">
+        <ProposedDateView timestamp={timestamp} />
+        <div className="column is-vcentered is-multiline is-mobile border-light-bottom is-flex is-full">
```
linxiao-li, line 28: Do we intend to have this div as a container or a child element or both?

packages/client/src/utils.js
```diff
@@ -132,3 +139,14 @@ export const parseIdentifier = (identifier) => {
...
   };
 };
+
+export const getSafeContactName = (contacts, address) => {
+  const contact = contacts.find((contact) => contact.address === address);
+  return contact?.name;
+};
+
+export const parseTimestamp = (timestamp) => {
+  console.log("timesta", timestamp)
```
linxiao-li, line 149: remove console log

packages/client/src/utils.js
```diff
@@ -132,3 +139,14 @@ export const parseIdentifier = (identifier) => {
...
+  console.log("time", new Date(parseInt(timestamp) * 1000).toLocaleString())
+  return new Date(parseInt(timestamp) * 1000).toLocaleString();
```
linxiao-li, line 151: I think we can bring in a third party library to handle the dates and time, check out Day.js https://www.skypack.dev/view/dayjs

packages/client/src/components/Actions/components/NFTView.js
```diff
@@ -0,0 +1,99 @@
...
+import { formatAddress, getSafeContactName, parseIdentifier } from "utils";
+import { useClipboard, useContacts } from "hooks";
+import { Web3Context } from "contexts/Web3";
+import Svg from "library/Svg";
+import ProposedDateView from "./ProposedDateView";
+
+const NFTView = ({ actionView, safeData }) => {
+  const { recipient, nftId, collectionId, timestamp } = actionView;
+  const { name: safeName, address: safeAddress } = safeData;
+  const { contractName: NFTName } = parseIdentifier(collectionId) || {};
```
linxiao-li, line 11: same here

packages/client/src/components/Actions/components/NFTView.js
```diff
@@ -0,0 +1,99 @@
...
+
+  const { getTreasuryNFTReference } = useContext(Web3Context);
+  const { contacts } = useContacts(safeAddress);
+  const [image, setImage] = useState();
+  const clipboard = useClipboard();
+
+  const { name: imageName, imageURI } = image || {};
```
linxiao-li, line 18: Same

packages/client/src/components/Actions/components/OwnerView.js
```diff
@@ -0,0 +1,34 @@
...
+
+const OwnerView = ({ actionView, safeData, isAdd }) => {
+  const { signerAddr, timestamp } = actionView;
+  const { address: safeAddress } = safeData;
+
+  const { contacts } = useContacts(safeAddress);
+  const signerName = getSafeContactName(contacts, signerAddr);
+
+  return (
+    <div className="mb-4">
```
linxiao-li, line 13: forgot to add class columns?

packages/client/src/components/Actions/components/TokenView.js
```diff
@@ -0,0 +1,66 @@
...
+              <span
+                className="pointer"
+                onClick={() => clipboard.copy(safeAddress)}
+              >
+                <Svg name="Copy" className="mt-1 ml-2 pointer" />
+              </span>
+            </div>
+          </div>
+        </div>
+        <div className="column is-vcentered is-multiline is-mobile border-light-bottom is-flex is-full">
```
linxiao-li, line 45: "Send From" and "Send To" rows are used and implemented the same way in both nft view and tokenview, consider extract into component and reuse

packages/client/src/pages/Safe.js
```diff
@@ -160,6 +160,36 @@ function Safe({ web3 }) {
     ).catch((error) => showErrorModal(error));
   };
 
+  const onRejectAction = async ({ uuid, intent }) => {
+    const latestBlock = await injectedProvider
```
linxiao-li, line 164: Any reason not to use the `createSignature` defined function in web3?

linxiao-li:
I'm seeing Invalid Date on modals
Screen Shot 2022-09-20 at 1 01 46 PM

packages/client/src/components/Actions/components/Confirmations.js
KsenijaZivkovic: As I understood from this document https://www.notion.so/Owner-Contacts-Logic-b5c5ee72eb5f4d4887d59c081d510ae4 - we should use saved addresses for logic in modals, send, approve/reject and confirmation
@amyliu6 please let me know your thoughts

jacksonConrad (approved):
Cadence Contract changes + scripts look good!
👍🏻 `getCurrentBlock().timestamp`

packages/client/src/components/Actions/components/Confirmations.js
KsenijaZivkovic: Thank you! I will update confirmations to contain names of the safe owners.
The information in the modal -> Sent from and Sent to -> i will leave it to be read from the safe contacts, if that's okay.

packages/client/src/pages/Safe.js
KsenijaZivkovic: There is a slight difference, because the message in approve/reject contains the uuid of the action, I'll try to consolidate those two into one function

packages/client/src/components/Actions/components/TokenView.js
KsenijaZivkovic: Can you please elaborate?

KsenijaZivkovic:
> I'm seeing Invalid Date on modals Screen Shot 2022-09-20 at 1 01 46 PM

Related to this one, the issue was that the contracts weren't upgraded - so we need to check if it's possible to just upgrade them or we need also to modify the contracts to V5

packages/client/src/components/Actions/components/TokenView.js
KsenijaZivkovic: Adding this class makes the elements properly aligned, but I can remove it and try an alternative way

linxiao-li: Doing an element inspection on columns class, it appears that columns does two things: 1. adds gap between the children elements, 2. set the container to `display:flex`. I doubt it has an impact on the alignment. Suggest adding `is-flex is-align-items-center` on the container div If the purpose here is to align the spans.

packages/client/src/components/Actions/components/TokenView.js
linxiao-li: `is-vcentered`, `is-multiline` and `is-mobile` are modifiers for the columns class, they need to be used in conjunction to the columns class(container class) but I'm seeing `column`, which is children element class, that's why I'm confused

almendili:
Looks great!

One small feedback for remove owner:
Update title: Remove safe owner > Remove Owner

mannynotfound (approved):
left some small comments but otherwise LGTM!

packages/client/src/App.sass
```diff
@@ -246,6 +250,10 @@ h2
 .button.is-small.with-icon>svg
   transform: scale(0.7)
 
+/********************* Custom utility css - Columns ********************/
+.column
```
linxiao-li, line 254: This will override the default column class, impacting every element with column class. Are we sure we want to do this? Is there an alternative way?

packages/client/src/App.sass
```diff
@@ -246,6 +250,10 @@ h2
...
+  padding-left: 0
```
mannynotfound, line 255: yeah we should probably just add `pl-0` wherever we dont want the default column padding since a lot of components use this class and could have unintended side effects

mannynotfound (changes requested):

packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -0,0 +1,65 @@
...
+          <OwnerView
+            actionView={actionView}
+            safeData={safeData}
+            isAdd={false}
+          />
+        );
+      case ACTION_TYPES.UPDATE_THRESHOLD:
+        return <ThresholdView actionView={actionView} safeData={safeData} />;
+      default:
+        return <></>;
```
linxiao-li, line 38:
```suggestion
        return null
```
reason: no need to render an empty div

packages/client/src/components/Actions/components/Confirmations.js
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <div className="p-4">
+      <div className="columns">
+        <span className="column">Confirmations</span>
+        <span className="column has-text-right">
+          {getNumberOfConfirmations()}
+        </span>
+      </div>
+      <div className="columns confirmations has-text-black">
```
linxiao-li, line 40: Children elements don't have column class attached, what are we trying to achieve with this columns class?

packages/client/src/App.sass
KsenijaZivkovic: You're right, reverting this!

packages/client/src/components/Actions/components/Confirmations.js
KsenijaZivkovic: Missed this one! I removed it and added display: flex to the confirmations class, because that was the only thing needed from the columns class

packages/client/src/utils.js
```diff
@@ -1,3 +1,4 @@
+import daysjs from "dayjs";
```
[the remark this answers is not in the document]
KsenijaZivkovic: Yup, just because it's small it's good enough :D plus it has a nice feature for formatting the date like in designs

mannynotfound (approved):

packages/client/src/App.sass
KsenijaZivkovic: removed this

linxiao-li (approved):

PR #109, opened 2022-09-20
reset default value of headerTitle when reusing modals

mannynotfound:
## Description

Fixes issue where once headerTitle is set, it gets appended to future modals because we're not clearing the modal state properly.

## Demo / Test Result

image

### DIFF
--- a/packages/client/src/contexts/Modal.js
+++ b/packages/client/src/contexts/Modal.js
@@ -49,6 +49,7 @@ const ModalProvider = ({ children }) => {
           closeOnBackgroundClick: true,
           showCloseButton: true,
           classNameModalContent: "",
+          headerTitle: "",
           ...customModalConfig,
         }));
       }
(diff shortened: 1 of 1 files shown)


KsenijaZivkovic:
I think this is not enough to solve this issue. 
If we go to the Safe Settings, in the screenshot you can see the behaviour:
- Open remove vault modal, and close it
- Open Add new owner modal
Screen Shot 2022-09-20 at 10 15 21 AM

Can you remove titles added with html from the modals - and add them via property headerText?

linxiao-li (commented):
LGTM

linxiao-li (approved):

mannynotfound:
@KsenijaZivkovic it seems like modals opened without `customModalConfig` wouldnt get their titles reset so they were still appearing incorrect. I fixed the code logic to always reset to default values, then override with `customModalConfig` if its provided. 

Will keep this PR focused on fixing the bug then will do a follow up PRs to get all of our modals using the same format

KsenijaZivkovic (approved):

PR #111, opened 2022-09-20
update TransactionList to new events/design

mannynotfound:
## Description

- add retrieval of `uuid` when getting treasuries so we can use that to query events
- update `refreshTreasury` / `fetchTreasury` to do a single request that fetches `uuid` + `signers` + `threshold`
- remove old transactions detail modal, going to implement new one in follow up PR

## Demo / Test Result

image


packages/client/src/components/TransactionList.js
```diff
@@ -190,17 +112,27 @@ function TransactionList({ safeData, transactions = [{ foo: "bar" }] }) {
     );
   };
 
-  return transactions.map((transaction, index) => {
-    return (
+  const transactionsToShow = [];
+  transactions.forEach(tx => {
+    // only show successes / rejections
+    const isTransfer = tx?.blockEventData?.actionView?.type.includes('Transfer');
+    const isRejection = tx?.flowEventId.includes('ActionDestroyed');
```
mannynotfound, line 119: currently the design only shows successful transfers or rejected actions so filtering out anything else

packages/client/src/components/TransactionList.js
```diff
@@ -134,50 +28,78 @@ function TransactionDetails({ safeOwners, transaction, onClose }) {
   );
 }
 
+function getTransferType(intent) {
+  if (intent.includes("from the treasury")) {
+    return "Sent";
+  }
```
mannynotfound, line 34: I don't love this but I'm not sure a better way to differentiate between in/out. It's not as easy as checking if `recipient === treasuryAddr` since the recipient could be the wallet address of the user that owns the treasury.

packages/client/src/components/TransactionList.js
```diff
@@ -1,78 +1,47 @@
...
+  status: 140,
+  date: 160,
+  action: 70,
+};
 
-function getDisplayStringForTokenTransfer(tokenTransfer) {
-  if (!tokenTransfer) {
-    return "";
-  }
+const Column = ({ className = "", children, style = {} }) => (
```
linxiao-li, line 12: Is this component purely for styling? What's the consideration of rendering by columns not rows?

packages/client/src/components/TransactionList.js
```diff
@@ -1,78 +1,47 @@
 import React from "react";
-import { useModalContext } from "../contexts";
-import { getFlowscanUrlForTransaction, shortenAddr } from "../utils";
+import { useModalContext } from "contexts";
 import TransactionStatusIcon from "./TransactionStatusIcon";
 
-const FALLBACK_RECIPIENT_NAME = "No name";
+const maxWidths = {
```
linxiao-li, line 5: We should avoid rendering with hardcode pixel values unless absolutely necessary. Anyway we can solve the problem with flex-box or bulma columns?

packages/client/src/components/TransactionList.js
```diff
@@ -1,78 +1,47 @@
...
-        className="has-background-black rounded-sm has-text-white is-flex is-align-items-center p-4 mr-2"
-      >
-        {shortenAddr(address)}
-      </div>
-    );
+function getUserFriendlyIntent(intent) {
+  const friendlyNameMap = {
+    "FlowToken": "FLOW",
   };
+  const newWords = intent.split(" ").map(word => {
```
linxiao-li, line 24: Can we move this to the utils?

packages/client/src/components/TransactionList.js
```diff
@@ -1,78 +1,47 @@
...
-        <div className={`has-text-${labelIsGrey ? "grey" : "black"}`}>
-          {value}
-        </div>
-      </div>
-    );
-  };
+  return newWords.join(" ");
+}
 
+const TransactionDetails = ({ transaction, onClose }) => {
```
linxiao-li, line 43: Can we move this into a separate file?

packages/client/src/components/TransactionList.js
```diff
@@ -132,52 +62,53 @@ function TransactionDetails({ safeOwners, transaction, onClose }) {
       </div>
     </>
   );
-}
+};
 
-function TransactionListItem({
+const TransactionListItem = ({
```
linxiao-li, line 67: Can we move to a separate file?

packages/client/src/components/TransactionList.js
```diff
@@ -132,52 +62,53 @@ function TransactionDetails({ safeOwners, transaction, onClose }) {
...
-      <div className="column has-text-grey">
-        {date.toLocaleTimeString("en-us")}
-      </div>
-      <div className="column has-text-right">
-        <span
-          className="is-underlined pointer"
-          onClick={() => onView(transaction)}
-        >
-          View
+    <div className={classes}>
```
linxiao-li, line 83: Seems the container div is missing "columns" class

packages/client/src/components/TransactionList.js
```diff
@@ -132,52 +62,53 @@ function TransactionDetails({ safeOwners, transaction, onClose }) {
...
+      <Column style={{ maxWidth: maxWidths.index }}>
+        {String(displayIndex).padStart(2, "0")}{' '}
```
linxiao-li, line 85: Can you help me understand what this part does?

packages/client/src/components/TransactionList.js
```diff
@@ -190,17 +121,35 @@ function TransactionList({ safeData, transactions = [{ foo: "bar" }] }) {
...
-  });
-}
+  // only show txs with intent
+  const transactionsToShow = transactions.filter(
+    tx => tx?.blockEventData?.actionView?.intent
+  );
+
+  return (
+    <div className="border-light rounded-sm">
+      <div className="is-flex has-text-grey border-light-bottom is-hidden-mobile">
```
linxiao-li, line 131: missing columns class?

packages/client/src/components/TransactionStatusIcon.js
```diff
@@ -3,17 +3,17 @@ import React from "react";
 const STATUS_TO_BACKGROUND_COLOR = {
   confirmed: "#18AA01",
   pending: "#FF8A00",
+  rejected: "#CD0B0B",
```
linxiao-li, line 6: @KsenijaZivkovic added a status Icon and colors in her pr, https://github.com/DapperCollectives/VESSEL/pull/108
I think this file can be deprecated now.

packages/client/src/hooks/useTreasuryTransactions.js
```diff
@@ -6,34 +6,17 @@ const NETWORK = process.env.REACT_APP_FLOW_ENV === "testnet" ? "test" : "main";
 const API_URL = `https://prod-${NETWORK}-net-dashboard-api.azurewebsites.net/api`;
 const FETCH_URL = `${API_URL}/company/${COMPANY_ID}/search?projectId=${PROJECT_ID}`;
```
linxiao-li, line 7: I still don't think it's a good idea to have configs hardcoded in the source code, I'll work on implementing a solution for the configs https://linear.app/dappercollectives/issue/VSL-214

linxiao-li (commented):
Looks like SafeTransaction and TransactionHistory is identical, any way we can combine and reuse?  Suggest breaking down TransactionList into separate files, and group TransactionHistory, TransactionList, TransactionDetails, TransactionItem into a folder to make files more organized.

packages/client/src/components/TransactionList.js
mannynotfound: was to reduce the boilerplate of `column is-flex is-align-items-center`. I guess we could rename `TransactionListItem` to `TransactionListRow` or simply `Row`

packages/client/src/components/TransactionList.js
mannynotfound: its necessary to define the widths here so that `intent` (Info column) can be take up all available width as the strings can be quite long.

packages/client/src/components/TransactionList.js
mannynotfound: so `columns` actually has all this extra baggage that makes it easier to just not use it in most cases... i agree its weird but adding it here means I have to do a bunch of overrides as well lol

image
image

packages/client/src/components/TransactionList.js
mannynotfound: this code was already here but as I understand it adds a zero to the beginning of the number until it reaches the desired length. This is so we show `01`, `02`, etc instead of `1`, `2,`

packages/client/src/components/TransactionList.js
mannynotfound: I think utils should be reserved for functions that are used in several places

packages/client/src/components/TransactionList.js
mannynotfound: this file is ~150 lines of code which is pretty average for our codebase, I think its fine to co-locate these as they are very coupled

packages/client/src/components/TransactionList.js
mannynotfound: I will be in a follow up PR! I have a ticket for just the detail modal

packages/client/src/components/TransactionList.js
linxiao-li: Would using flex-grow help?

packages/client/src/components/TransactionList.js
linxiao-li: Yeah and looks like only 2 out of 5 columns need `is-flex is-align-items-center`

packages/client/src/components/TransactionList.js
linxiao-li: The function is to display the intent in a more friendly way, I can see this could be used in other places.

packages/client/src/hooks/useTreasuryTransactions.js
mannynotfound: we could make `REACT_APP_GRAFFLE_API_URL_TESTNET` and `REACT_APP_GRAFFLE_API_URL_MAINNET` and then switch which one we use based on `NETWORK` but its just more code and this shouldn't need to change / be configured that often

packages/client/src/components/TransactionList.js
linxiao-li: I remember having a similar discussion before, it is true that those two components are very closely coupled, the problem is average length of file and how closely components are coupled are subjective, not as easy to execute on the simple rule of one component per file. As the codebase grows bigger and more complex, sometimes we need these cookie cutter rules to keep the codebase maintainable. @amyliu6 What's your opinion?

packages/client/src/components/TransactionList.js
linxiao-li: I see, would `columns is-gapless` help solving the problem?

mannynotfound:
@amyliu6 updated to actual latest figma designs!

image

@linxiao-li renamed `TransactionList` => `TransactionTable` since that is more of a table and renamed `TransactionListItem` to `Row` to make more sense with `Column`. Also refactored `SafeTransactions`/`TransactionHistory` such that now we only have the code for fetching treasury events in `SafeTransactions` and `TransactionHistory` just imports `SafeTransactions`

packages/client/src/hooks/useTreasuryTransactions.js
linxiao-li: I'm thinking more like a `useConfig` hook, to hide all these implementation details of string concatenation

packages/client/src/components/TransactionStatusIcon.js
mannynotfound: would rather update this in a follow up PR after #108 is merged to not be dependent on it, or if it gets merged before this PR does I can update

packages/client/src/components/TransactionList.js
mannynotfound: that adds this and a few other styles that are troublesome 

image

The reason for not using bulma `columns` is because it assumes things of equal width or in neat little fractions like 1/3 etc but really this is a table where most of the columns are always going to be same width except for the variable string "Info" column.

packages/client/src/components/TransactionList.js
mannynotfound: I would vote to follow CAST's eslint rules so we have parity across the organization: https://github.com/DapperCollectives/CAST/blob/main/frontend/packages/client/package.json#L80-L85

linxiao-li: That's perfect! I'll make sure to add this rule to our eslint config

packages/client/src/components/TransactionStatusIcon.js
KsenijaZivkovic, line 6: Can we use here the colors defined in App.sass instead?
We have success, warning and danger classes for it

packages/client/src/components/TransactionTable.js
```diff
@@ -0,0 +1,134 @@
...
+        onClose={() => modalContext.closeModal()}
+      />
+    );
+  };
+
+  // only show txs with intent
+  const transactionsToShow = transactions.filter((tx) => tx?.blockEventData?.actionView?.intent);
+
+  return (
+    <div className={`border-light rounded-sm ${className}`}>
```
KsenijaZivkovic, line 113: We can use the class .table-border to add to rows border light bottom

packages/client/src/components/TransactionTable.js
```diff
@@ -0,0 +1,134 @@
...
+const Column = ({ className = "", children, style = {} }) => (
+  <div className={`column is-flex is-align-items-center ${className}`} style={style}>
+    {children}
+  </div>
+);
+
+const Row = ({ transaction, displayIndex, isLastInList, onView }) => {
+  const classes = [
+    "py-4 is-flex is-align-items-center is-justify-content-space-between",
+    isLastInList ? "" : "border-light-bottom",
```
KsenijaZivkovic, line 67: We don't need the prop isLastInList, we can use class .table-border for parent element

packages/client/src/components/SafeHome.js
```diff
@@ -3,24 +3,11 @@ import SafeOverview from "./SafeOverview";
...
-    <SafeOverview allBalance={allBalance} allNFTs={allNFTs} />
-    <ThingsToDo
-      actions={actions}
-      onSign={onSign}
-      onConfirm={onConfirm}
-      safeData={safeData}
-    />
-    <TransactionHistory safeData={safeData} address={address} />
+    <SafeOverview {...props} />
+    <ThingsToDo {...props} />
```
mannynotfound, line 9: `ThingsToDo` was missing `onReject` and `onApprove` so wasnt able to take those actions on items. since `SafeHome` is just a wrapper component that bundles other components to render, lets just spread all props to its children and children can decide what to use

linxiao-li: Pasting it here again in case you missed it:
I would suggest against passing props to all children
1. It creates redundant dependencies, causing components to be dependent and to re-render on irrelevant state changes.
2. It hurts readability, in order to figure out what's been passed to the component, another developer has to to a lot of tracking and code tracing.

onReject and onApprove is missing because of a bad merge conflict, creating ways to accommodate it will lead to more error in the future

mannynotfound: just trying to reduce boilerplate and places that need to be updated, but can update to xenias solution

mannynotfound:
@linxiao-li since xenia's PR merge I moved from `getFriendlyIntent` => `utils.formatActionString` and `TransactionStatusIcon` => `Svg name="Status"` 🎉 also just removed `column` and added the styles from the column I needed

@amyliu6 here's what mobile looks like, the "header" goes away and some of the columns are also hidden

image

KsenijaZivkovic:
Looks good! One small thing: border seems to shrink when we emulate this on the mobile view.
Screen Shot 2022-09-26 at 11 01 51 AM

packages/client/src/pages/Safe.js
```diff
@@ -62,6 +62,18 @@ const ReceiveTokens = ({ name, address }) => {
   );
 };
 
+const EmptyTransactions = () => (
```
linxiao-li, line 65: We have a `library/EmptyTableWithCTA` component you can modify and use

mannynotfound:
> Looks good! One small thing: border seems to shrink when we emulate this on the mobile view. Screen Shot 2022-09-26 at 11 01 51 AM

seems like this is coming from an unrelated style - we should probably fix this in a separate PR as it affects a lot of things
image

packages/client/src/App.sass
```diff
@@ -412,6 +412,9 @@ input:focus-visible
   flex: 2 0 0
 .flex-3
   flex: 3 0 0
+.flex-9
+  flex: 9 0 0
```
linxiao-li, line 416: Should probably set to flex: 9 9 9 (or flex:9 in short) if we want this column take 9 spaces on all screen sizes

KsenijaZivkovic (approved):

linxiao-li (approved):

PR #116, opened 2022-09-23
Reverted change 

KsenijaZivkovic:
## Description

Returned onApprove, onReject functions to enable signing/rejecting the actions
Updates execution limits to enable transactions on testnet

### DIFF
--- a/packages/client/src/components/SafeHome.js
+++ b/packages/client/src/components/SafeHome.js
@@ -9,14 +9,16 @@ const SafeHome = ({
   allNFTs,
   actions,
   address,
-  onSign,
+  onApprove,
+  onReject,
   onConfirm,
 }) => (
   <>
     <SafeOverview allBalance={allBalance} allNFTs={allNFTs} />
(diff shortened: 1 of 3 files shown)


amyliu6 (approved):

mannynotfound (approved):
I ran into this on my branch as well, I fixed it by just passing `props` to all the children of `SafeHome` since there is no business logic in SafeHome and that way we dont have to explicitly pass every prop to its children https://github.com/DapperCollectives/VESSEL/pull/111/commits/3abfecd79a3b1256267356af1fa20468b26781b0#r978989801

linxiao-li (approved):

linxiao-li:
@mannynotfound I would suggest against passing props to all children
1. It creates redundant dependencies, causing components to be dependent and to re-render on irrelevant state changes.
2. It hurts readability, in order to figure out what's been passed to the component, another developer has to to a lot of tracking and code tracing.

PR #118, opened 2022-09-24
VSL-187: Updated status and actions in Things To Do

KsenijaZivkovic:
## Description

What is done:
- Renamed onConfirm to onExecute
- Renamed statuses and buttons, and added proper Svgs
- Added titles above the things to do table

NOTE: Parts of the text weren't bolded because it would require a lot of text manipulation, which is prone to errors and constant modifications in case of intent changes. This text comes from BE (action view -> intent).

## Demo / Test Result

Screen Shot 2022-09-28 at 3 40 48 PM

Screen Shot 2022-09-28 at 3 41 06 PM

Screen Shot 2022-09-28 at 3 41 33 PM

Screen Shot 2022-09-28 at 3 41 55 PM


mannynotfound:
empty/mobile views are a bit off:

image

image

packages/client/src/components/Actions/ActionsList.js
```diff
@@ -54,7 +55,7 @@ function ActionsList({
   if (!actions.length) {
     ActionComponents.push(
       <div
-        className="column is-full p-0 is-flex is-justify-content-center is-align-items-center has-text-grey"
+        className="p-0 is-justify-content-center is-align-items-center has-text-grey"
```
linxiao-li, line 58:
```suggestion
        className="p-0 is-flex is-justify-content-center is-align-items-center has-text-grey"
```

packages/client/src/components/Actions/ActionsList.js
```diff
@@ -63,53 +64,55 @@ function ActionsList({
...
+      const actionPrompt = executionReady
+        ? "Pending Execution"
+        : "Pending Confirmation";
+      const actionCopy = executionReady ? "Execute" : "Confirm";
+      const actionFn = executionReady ? onExecute : openApproveOrRejectModal;
       ActionComponents.push(
         <div
           key={action.uuid}
-          className={`p-5 ${borderClass} is-flex column is-full`}
+          className="p-5 m-0 columns is-flex is-align-items-center"
```
linxiao-li, line 82:
```suggestion
          className="p-5 m-0 columns is-align-items-center"
```

packages/client/src/library/Svg/Confirm.js
```diff
@@ -0,0 +1,19 @@
+const Confirm = ({ width = "17", height = "16", className = "" }) => (
```
linxiao-li, line 1: Can we rename this svg to describe the icon (for example, CircleCheck)? It's easier to be referenced by other components. Same with Execute

KsenijaZivkovic: Not sure which other components would use them, it looks like it would be used only for confirming/executing, plus these icons might be changed so I would leave the name as-is for now, and change it when we get the final design.

KsenijaZivkovic: Updated the names to match the look of the icon!

packages/client/src/components/Actions/ActionsList.js
KsenijaZivkovic: Fixed

packages/client/src/components/Actions/ActionsList.js
KsenijaZivkovic: Fixed

KsenijaZivkovic:
> empty/mobile views are a bit off:
> 
> image
> 
> image

@mannynotfound fixed this too!

packages/client/src/components/Actions/ActionsList.js
```diff
@@ -53,63 +54,66 @@ function ActionsList({
...
     );
   } else {
+    ActionComponents.push(
+      <div className="px-5 m-0 columns has-text-left has-text-grey is-align-items-center is-hidden-touch">
+        <div className="column px-0 is-1">#</div>
+        <div className="column px-0 is-4">Info</div>
+        <div className="column is-3">Status</div>
+        <div className="column px-0 is-2">Signatures</div>
+        <div className="column px-0 is-2">Action</div>
+      </div>
```
mannynotfound, line 69: can we make this a `table` and use the markup structure from https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/components/TransactionTable.js#L73-L93 ?

since both of these are on the same page they should be nearly identical in spacing/mobile 

image

image

KsenijaZivkovic: In case of mobile - confirm button should be below, is that possible to achieve with tables?

mannynotfound: @KsenijaZivkovic might not be - you might just have to make the desktop confirm `is-hidden-mobile` and then have a "mobile" confirm button that is just a full table row only visible on mobile

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,137 @@
...
+  const executionReady = totalSigned >= safeData.threshold;
+  const background = executionReady ? "has-text-danger" : "has-text-warning";
+  const actionPrompt = executionReady
+    ? "Pending Signature"
+    : "Pending Approval";
+  const actionText = executionReady ? "Sign" : "Approve";
+  const actionSvg = executionReady ? "Quill" : "EmptyCheck";
+  const actionFn = executionReady ? onExecute : openApproveOrRejectModal;
+
+  const { getActionView } = useContext(Web3Context);
```
linxiao-li, line 61: Line 61 62 63 should be placed on top before they are used. ESlint should be able to catch that, let me know if you are not seeing the warning from ESLint

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,137 @@
...
+  const { showErrorModal } = useErrorMessage();
+  const { openModal, closeModal } = useModalContext();
+
+  return (
+    <>
+      <tr className="py-4 is-flex is-align-items-center">
+        <td className="p-3 flex-1">{String(displayIndex).padStart(2, "0")}</td>
+        <td className="p-3 flex-7">{formatActionString(action.intent)}</td>
+        <td className="p-3 flex-4 is-hidden-touch">
+          <Svg key={background} name="Status" className={background} />
```
linxiao-li, line 71: what does the key property do?

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,137 @@
...
+  const actionText = executionReady ? "Sign" : "Approve";
+  const actionSvg = executionReady ? "Quill" : "EmptyCheck";
+  const actionFn = executionReady ? onExecute : openApproveOrRejectModal;
+
+  const { getActionView } = useContext(Web3Context);
+  const { showErrorModal } = useErrorMessage();
+  const { openModal, closeModal } = useModalContext();
+
+  return (
+    <>
```
linxiao-li, line 66: Looks like we can remove this empty div and just return tr

packages/client/src/pages/Safe.js
```diff
@@ -29,20 +29,33 @@ const ReceiveTokens = ({ name, address }) => {
...
-          <QRCode className="ml-5" value={`https://flowscan.org/account/${address}`} />
+          <QRCode
+            className="ml-5"
+            value={`https://flowscan.org/account/${address}`}
+          />
         </div>
-        <button className="button is-transparent mt-5" onClick={() => clipboard.copy(address)}>
-          {clipboard.textJustCopied === address ? "Copied" : "Copy Safe Address"}
+        <button
+          className="button is-transparent mt-5"
```
linxiao-li, line 46: missing type property

packages/client/src/components/Actions/ActionsListTable.js
KsenijaZivkovic: There was an issue - color didn't change for the svg once signer approved the action, so when key is changed the color is updated as well

packages/client/src/components/Actions/ActionsListTable.js
KsenijaZivkovic: There are two <tr> tags inside, so it should stay

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,137 @@
+import { useContext } from "react";
+import { useErrorMessage } from "hooks";
+import { formatActionString } from "utils";
+import { useModalContext } from "contexts";
+import { Web3Context } from "contexts/Web3";
```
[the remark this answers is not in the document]
KsenijaZivkovic: Doesn't work :(

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,137 @@
...
+
+const Row = ({
+  safeData,
+  action,
+  displayIndex,
+  onApprove,
+  onReject,
+  onExecute,
+}) => {
+  const onApproveAction = async (action) => {
```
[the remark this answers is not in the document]
KsenijaZivkovic: Updated!

packages/client/src/components/Actions/ActionsListTable.js
KsenijaZivkovic: Fixed

packages/client/src/components/Actions/ActionsListTable.js
linxiao-li: Got it, I'll fix the Svg loader

packages/client/src/components/Actions/ActionsListTable.js
```diff
@@ -0,0 +1,135 @@
...
+        confirmations={signerResponses}
+        onApprove={() => onApproveAction(action)}
+        onReject={() => onRejectAction(action)}
+      />,
+      { headerTitle: "Approval Required" }
+    );
+  };
+
+  return (
+    <table className={`border-light rounded-sm`}>
```
linxiao-li, line 109: Looks like we don't need the template literal here

linxiao-li (approved):
LGTM overall!

packages/client/src/components/Actions/ActionsListTable.js
KsenijaZivkovic: The second row is visible only in mobile view, to display the button, and in the first row the button is visible in desktop view

KsenijaZivkovic: I used Manny's idea from this comment https://github.com/DapperCollectives/VESSEL/pull/118#discussion_r982906870

packages/client/src/components/Actions/ActionsListTable.js
KsenijaZivkovic: Fixed!

mannynotfound (approved):

almendili:
Front-end feedback:
- Can we use title case for "Transaction history" -> "Transaction History"
- Can we add a grey line between items in Things To Do and Transaction History
- Can we reduce the padding for each line. Right I think it's 1rem. Could we go to .5rem

Otherwise, everything looks great!

PR #122, opened 2022-09-27
add transaction details modal

mannynotfound:
## Description

- move `components/Actions/components/Confirmations` => `components/Confirmations` as its used in multiple places now
- create `getStatusColor` util function to get css colors based on transaction status
- modify `TransactionTable` to take the events data from graffle and make it more like the data our components are used to rendering
- move `TransactionSuccessModal` => `modals/Transaction/Success` 
- create `modals/Transaction/Details`
- create `modals/Transaction/BannerInfo` to be used by both `modals/Transaction/Success` and `modals/Transaction/Details`

## Demo / Test Result

image
image


almendili:
This look great!

A few comments on the FE:
- Can we switch the "external link" icon color in the Flowscan button to the same purple as the text.
- Switch "approved" to "Approved"
- Switch "Confirmations" to "Signatures" (we just updated this in the design)
- In the Approved card, can we change the font size for where it says "1" and the top line with address to 1.875rem or 30px.
- When we only have an address (no saved names), can we skip the 2nd line that repeats the address.

packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -3,15 +3,9 @@ import NFTView from "./NFTView";
...
+import Confirmations from "components/Confirmations";
 
-const ActionRequired = ({
-  safeData,
-  actionView,
-  confirmations,
-  onReject,
-  onApprove,
-}) => {
+const ActionRequired = ({ safeData, actionView, confirmations, onReject, onApprove }) => {
```
amyliu6, line 8: @mannynotfound  please help cross check the formatting, seems the files are committed with a different format, which will result unrelated diff in PR review.

mannynotfound: @amyliu6 this is happening in my editor since our prettier rules were updated - I think its expected we use this setting so any changes to old files will have unrelated prettier changes:

image

I agree it makes PR a bit harder to reason about but hopefully its just a growing pain to a better future 😅

mannynotfound: @amyliu6 actually I was just on a new editor that was missing https://github.com/prettier/prettier-vscode#prettierrequireconfig-default-false 😓 I've updated PR now so format should be the same.

amyliu6: cool. we also have pre-commit hook to auto format during commit, curious why that didn't kick-in? 🤔

packages/client/src/modals/Transaction/Details.js
```diff
@@ -0,0 +1,111 @@
...
+                  <span className="pointer" onClick={() => clipboard.copy(recipient)}>
+                    <Svg name="Copy" className="mt-1 ml-2 pointer" />
+                  </span>
+                </div>
+              </div>
+            </div>
+          </>
+        )}
+        {(actionType === ACTION_TYPES.ADD_OWNER_UPDATE_THRESHOLD ||
+          actionType === ACTION_TYPES.REMOVE_OWNER_UPDATE_THRESHOLD) && (
```
amyliu6, line 81: there should be actionType to `updateThreshold` only right? is that one missing here?

mannynotfound: @amyliu6 you're right! I was also using outdated action type variable names so ive added `UPDATE_THRESHOLD` and corrected the old variable names to current main branch names

KsenijaZivkovic (approved):

packages/client/src/components/TransactionTable.js
```diff
@@ -58,25 +33,50 @@ const Row = ({ transaction, displayIndex, onView }) => {
...
+    });
+    openModal(
+      <TransactionDetailsModal
+        onClose={closeModal}
+        safeData={safeData}
+        // massage data from graffle to work with TransactionDetails
+        transaction={{
+          status: transaction.status,
+          txID: transaction.flowTransactionId,
+          date: transaction.eventDate,
```
linxiao-li, line 58: prefer destructure before
```suggestion
 const {status, flowTransactionId, eventDate} = transaction
 ...
          status: status,
          txID: flowTransactionId,
          date: eventDate,
```

packages/client/src/modals/Transaction/Details.js
```diff
@@ -0,0 +1,114 @@
...
+import { ACTION_TYPES } from "constants/enums";
+
+const TransactionDetails = ({ onClose, safeData = {}, transaction = {} }) => {
+  const { recipient, date, status, signers } = transaction;
+  const clipboard = useClipboard();
+  const statusColor = getStatusColor(status);
+  const { contacts } = useContacts(safeData.address);
+  const actionType = transaction.type;
+
+  console.log(actionType)
```
linxiao-li, line 16: remove console

packages/client/src/components/TransactionTable.js
```diff
@@ -58,25 +33,50 @@ const Row = ({ transaction, displayIndex, onView }) => {
...
+          signers,
```
linxiao-li, line 59: Looks like the consumers of this signers object use it with Object.keys, consider using an Array instead of an object, it will help simplify the code

packages/client/src/modals/Transaction/BannerInfo.js
```diff
@@ -0,0 +1,86 @@
+import React, { useEffect, useContext, useState } from "react";
+import { Web3Context } from "contexts/Web3";
+import { ACTION_TYPES } from "constants/enums";
+import Svg from "library/Svg";
+import { formatAddress, getTokenMeta, parseIdentifier, getNameByAddress } from "utils";
+
+const BannerInfo = ({ className = "", actionData = {}, contacts = {}, signers = [] }) => {
```
linxiao-li, line 7: Didn't see any place passing the signers property to BannerInfo, is it being used?

packages/client/src/modals/Transaction/BannerInfo.js
```diff
@@ -0,0 +1,86 @@
...
+import Svg from "library/Svg";
+import { formatAddress, getTokenMeta, parseIdentifier, getNameByAddress } from "utils";
+
+const BannerInfo = ({ className = "", actionData = {}, contacts = {}, signers = [] }) => {
+  const { getAccountNFTReference } = useContext(Web3Context);
+  const [image, setImage] = useState();
+  const { recipient, nftId, collectionId, vaultId, tokenAmount, signerAddr } = actionData;
+  const { displayName, tokenType } = getTokenMeta(vaultId);
+  const { contractName: NFTName, contractAddress: NFTAddress } = parseIdentifier(collectionId);
+  const { name: imageName, imageURI } = image || {};
```
linxiao-li, line 13: Suggest moving this part to line 9 on initializing the state so the structure is clearer
```suggestion
  const [image, setImage] = useState({
      imageName:undefined,
      imageURI:undefined
  })

```

packages/client/src/modals/Transaction/BannerInfo.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const contactName = getNameByAddress(contacts, signerAddr);
+
+  return (
+    <div className={className}>
+      {actionType === ACTION_TYPES.TRANSFER_NFT && (
+        <>
+          {imageURI && (
+            <img
+              className="columns is-vcentered is-multiline is-mobile mr-2 mt-2 success-modal-image"
```
linxiao-li, line 39: columns class should be attached to a container dom but this img doesn't have any child, what is the effect we  are trying to achieve here?

packages/client/src/modals/Transaction/BannerInfo.js
```diff
@@ -0,0 +1,86 @@
...
+              src={imageURI}
+              alt={imageName}
+            />
+          )}
+          <span className="columns is-vcentered is-multiline is-mobile mr-2 is-size-4 is-family-monospace">
```
linxiao-li, line 44: Same here

packages/client/src/modals/Transaction/Details.js
```diff
@@ -0,0 +1,114 @@
...
+          </div>
+        </div>
+        {(actionType === ACTION_TYPES.TRANSFER_TOKEN ||
+          actionType === ACTION_TYPES.TRANSFER_NFT) && (
+          <>
+            <div className="border-light-top is-flex py-4">
+              <div className="flex-1 has-text-grey">Sent from</div>
+              <div className="flex-1">
+                <div>
+                  <b>{safeData.name}</b>
```
linxiao-li, line 59: prefer use destructuring on the properties of safeData

packages/client/src/modals/Transaction/Details.js
```diff
@@ -0,0 +1,114 @@
...
+          actionType === ACTION_TYPES.UPDATE_THRESHOLD) && (
+          <div className="border-light-top is-flex py-4">
+            <div className="flex-1 has-text-grey">Signature Threshold</div>
+            <div className="flex-1 has-text-weight-bold">
+              {Object.keys(signers).length} of {safeData.threshold} owners
+            </div>
+          </div>
+        )}
+        <div className="border-light-top pb-4 has-text-grey">
+          <Signatures safeData={safeData} confirmations={signers} />
```
linxiao-li, line 93: Looks like Signatures only needs safeData.safeOwners, can we just pass safeOwners instead of the whole safeData?

mannynotfound:
was able to test the NFT transfer flow:

image

packages/client/src/modals/Transaction/BannerInfo.js
mannynotfound: ah yeah it was missing being passed down from `Details.js`, added it!

packages/client/src/modals/Transaction/BannerInfo.js
mannynotfound: I copy/pasted this code from https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/modals/TransactionSuccessModal.js#L63-L79

was just trying to change as little as possible in moving stuff around

packages/client/src/components/TransactionTable.js
mannynotfound: I was just trying to match the format already used by https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/src/components/Actions/components/Confirmations.js#L8-L14 so was trying to change as little as possible.

linxiao-li (approved):

PR #123, opened 2022-09-28
VSL-120: Updated modal for signers and threshold updates

KsenijaZivkovic:
## Description

After merging add/remove owner with update of the threshold, also the modals were updated to match this change

## Demo / Test Result

Screen Shot 2022-09-27 at 8 27 36 PM
Screen Shot 2022-09-27 at 8 28 35 PM

### DIFF
--- a/packages/client/src/components/Actions/components/ActionRequired.js
+++ b/packages/client/src/components/Actions/components/ActionRequired.js
@@ -1,7 +1,7 @@
 import { ACTION_TYPES } from "constants/enums";
 import NFTView from "./NFTView";
 import TokenView from "./TokenView";
-import OwnerView from "./OwnerView";
+import SignerThresholdView from "./SignerThresholdView";
 import ThresholdView from "./ThresholdView";
 import Confirmations from "./Confirmations";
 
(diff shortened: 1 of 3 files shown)


packages/client/src/components/Actions/components/ActionRequired.js
```diff
@@ -20,13 +20,17 @@ const ActionRequired = ({
         return <TokenView actionView={actionView} safeData={safeData} />;
       case ACTION_TYPES.TRANSFER_NFT:
         return <NFTView actionView={actionView} safeData={safeData} />;
-      case ACTION_TYPES.ADD_OWNER:
+      case ACTION_TYPES.ADD_SIGNER_UPDATE_THRESHOLD:
```
mannynotfound, line 23: is it still possible to trigger just `ADD_OWNER` ? if so maybe this should be 

```
case ACTION_TYPES.ADD_SIGNER_UPDATE_THRESHOLD:
case ACTION_TYPES.ADD_OWNER:
```

mannynotfound (commented):
code LGTM, just wondering if maybe this should be in `src/modals`? Some of this functionality seems duplicated in `TransactionSuccessModal` as well as my PR https://github.com/DapperCollectives/VESSEL/pull/122

packages/client/src/components/Actions/components/ActionRequired.js
KsenijaZivkovic: From the UI it's not possible, but I see that it still exists in the contracts.
@jacksonConrad do we intend to remove AddSigner and RemoveSigner actions from the TreasuryActions.cdc?

KsenijaZivkovic:
> code LGTM, just wondering if maybe this should be in `src/modals`? Some of this functionality seems duplicated in `TransactionSuccessModal` as well as my PR #122

Yeah, this is a tough one to handle here, since we have 3 PRs that are based on the same code, we can chat tomorrow and see how to separate this into reusable components. I would like to tackle this in a different ticket if that sounds okay.

linxiao-li (approved):
LGTM

mannynotfound (approved):

PR #120, opened 2022-09-26
vsl-219: Update UI for Send Modal

linxiao-li:
## Description
1. Refactor send modal to align with figma design
2. Refactor dropdown to handle more functions
3. Create Tooltip reusable component

## Demo / Test Result
- Send token flow

https://user-images.githubusercontent.com/7423845/193361946-6c91197f-65fd-4e2f-a676-3cfee907bba2.mov

- Send nft flow

https://user-images.githubusercontent.com/7423845/193361968-bf0b8766-273b-4d47-81d7-8d0efb45a33e.mov

- Dropdown and Tooltip
- Dropdown can search by address, owner name
- Dropdown input box can be used as address input

https://user-images.githubusercontent.com/7423845/193362213-5e741e8e-4829-4eae-854d-241588cdf106.mov

PR #125, opened 2022-10-03
update any instance of address validation to use InputAddress

mannynotfound:
## Description

We have several places in our app where we have an input for FLOW addresses which we want to validate against the FLOW blockchain and sometimes run additional validation logic on it. This PR moves all those custom `<input>` to use our shared `<InputAddress>` so styling is consistent and to minimize places to update.

## Demo / Test Result

image
image
image


packages/client/src/library/components/InputAddress.js
```diff
@@ -4,7 +4,14 @@ import { useAddressValidation } from "hooks";
...
 import { isAddr, formatAddress } from "utils";
 
-const InputAddress = ({ web3, value, isValid, onChange }) => {
+const InputAddress = ({ 
+  web3, 
+  value, 
+  isValid, 
+  onChange, 
+  className = "", 
+  readOnly = false,
```
mannynotfound, line 13: add `className` to add classes needed for situational positioning/styling and `readOnly` to render a disabled input

packages/client/src/library/components/InputAddress.js
```diff
@@ -18,17 +25,19 @@ const InputAddress = ({ web3, value, isValid, onChange }) => {
   };
 
   return (
-    <div className="is-flex">
+    <div className={`is-flex ${className}`}>
       <div className="flex-1" style={{ position: "relative" }}>
         <input
-          style={{ height: 48 }}
-          className="border-light rounded-sm column is-full p-2 mt-2"
+          className="border-light rounded-sm column is-full p-4"
```
mannynotfound, line 31: moved `p-2` => `p-4` to match designs closer / be consistent throughout app

packages/client/src/library/components/InputAddress.js
```diff
@@ -18,17 +25,19 @@ const InputAddress = ({ web3, value, isValid, onChange }) => {
...
           type="text"
           value={value}
           onChange={onValueChange}
+          disabled={readOnly}
         />
         {isValid && (
-          <div style={{ position: "absolute", right: 17, top: 20 }}>
+          <div 
+            className="is-flex is-align-items-center" 
+            style={{ position: "absolute", right: 15, top: 0, height: "100%" }}>
```
mannynotfound, line 40: use a container with 100% height thats `align-items: center` which should be more flexible than fixed absolute positioning

packages/client/src/components/SafeSettings/components/Owners/AddSafeOwner.js
```diff
@@ -49,22 +42,12 @@ const AddSafeOwner = ({ treasury, safeOwners }) => {
...
-            {addressValid && (
-              <div
-                style={{ position: "absolute", right: 17, top: 19 }}
-              >
-                <Svg name="Check" height="15" width="15" />
-              </div>
-            )}
-          </div>
+          <InputAddress
+            className={addressValidClass}
```
KsenijaZivkovic, line 46: It looks like this addressValidClass depends if the address is valid, so that logic can be moved to InputAddress?
This way we can ensure all the inputs have is-success/is-error class attached

mannynotfound: @KsenijaZivkovic good call - added!

jacksonConrad:
Everything looks good, although I noticed that on your branch i'm unable to click `Create Safe`, despite having the required fields populated.

Screen Shot 2022-10-03 at 4 04 44 PM

packages/client/src/components/SafeOwners.js
```diff
@@ -13,9 +13,10 @@ function SafeOwners({
     setSafeOwners([...newOwners]);
   };
 
-  const onOwnerAddressChange = (value, idx) => {
+  const onOwnerAddressChange = (value, isValid, idx) => {
     const newOwners = safeOwners.slice(0);
     newOwners[idx].address = value;
+    newOwners[idx].isValid = isValid;
     setSafeOwners([...newOwners]);
   };
```
linxiao-li, line 21: Prefer following the style guide
```suggestion
  const onOwnerAddressChange = (value, isValid, idx) => {
    const newOwners = [...safeOwners]
    newOwners[idx].address = value;
    newOwners[idx].isValid = isValid;
    setSafeOwners(newOwners);
  };
```

packages/client/src/components/SendTokens/components/AddressInput.js
```diff
@@ -28,23 +19,12 @@ const AddressInput = () => {
...
-            onChange={onrecipientChange}
-          />
-          {recipientValid && (
-            <div style={{ position: "absolute", right: 17, top: 20 }}>
-              <Svg name="Check" />
-            </div>
-          )}
-        </div>
-      </div>
+      <InputAddress
```
linxiao-li, line 22: This input Address component is being replaced by addressDropdown. I think we can safely delete this file now https://github.com/DapperCollectives/VESSEL/pull/120/files#diff-e8e43eadee51d689965c6c2bd4a914f301aea66e12351d018baa6cd3567d1422

packages/client/src/library/components/InputAddress.js
```diff
@@ -4,7 +4,14 @@ import { useAddressValidation } from "hooks";
 import Svg from "library/Svg";
 import { isAddr, formatAddress } from "utils";
 
-const InputAddress = ({ web3, value, isValid, onChange }) => {
+const InputAddress = ({ 
+  web3,
```
linxiao-li, line 8: Do we still need web3 as a prop when we expect isValid and onChange to handle the validation?

mannynotfound:
> Everything looks good, although I noticed that on your branch i'm unable to click `Create Safe`, despite having the required fields populated.
> 
> Screen Shot 2022-10-03 at 4 04 44 PM

Small bug, should be updated now!

PR #126, opened 2022-10-03
add prettier setup from CAST

mannynotfound:
## Description

Borrow the `prettier` setup from [CAST](https://github.com/DapperCollectives/CAST/blob/main/frontend/packages/client/.prettierrc) so that we can easily format our files while developing and correct small issues automatically with `eslint --fix` + `prettier --write`.

### DIFF
--- a/packages/client/.eslintrc.json
+++ b/packages/client/.eslintrc.json
@@ -4,7 +4,11 @@
     "es2021": true,
     "node": true
   },
-  "extends": ["plugin:react/recommended", "airbnb"],
+  "extends": [
+    "plugin:react/recommended",
+    "plugin:prettier/recommended",
+    "airbnb"
+  ],
   "overrides": [],
   "settings": {
     "import/resolver": {
@@ -24,7 +28,7 @@
       }
     }
   },
-  "plugins": ["react", "@babel"],
+  "plugins": ["react", "@babel", "prettier"],
   "rules": {
     "indent": ["off"],
     "prefer-const": ["error"],
@@ -50,7 +54,18 @@
         "ignoreStateless": true
       }
     ],
-    "jsx-a11y/click-events-have-key-events": ["off"]
+    "jsx-a11y/click-events-have-key-events": ["off"],
+    "prettier/prettier": [
+      "error",
+      {
+        "singleQuote": true,
+        "comma-dangle": ["error", "always-multiline"]
+      },
+      {
+        "usePrettierrc": true
+      }
+    ],
+    "import/order": [0]
   },
   "ignorePatterns": ["build/**/*", "public/**/*", "src/flow/**/*"]
 }
--- a/packages/client/.prettierrc
+++ b/packages/client/.prettierrc
@@ -3,7 +3,7 @@
   "singleQuote": true,
   "printWidth": 80,
   "tabWidth": 2,
-  "tabs": false,
+  "useTabs": false,
   "semi": true,
   "bracketSpacing": true,
   "importOrder": [
--- a/packages/client/.prettierrc
+++ b/packages/client/.prettierrc
@@ -1,0 +1,22 @@
+{
+  "trailingComma": "es5",
+  "singleQuote": true,
+  "printWidth": 80,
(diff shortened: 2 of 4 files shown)


jacksonConrad (approved):

KsenijaZivkovic:
What are the steps to enable these commands to work? I get the errors: command not found: eslint/prettier
Do we need to have it globally installed?
Is it possible to enable the pre-commit hook to perform the fix/formatting too?

linxiao-li (approved):

mannynotfound:
@KsenijaZivkovic which command were you trying? the following should work after a `yarn install` (I only have `yarn` and `create-react-app` installed globally)

I was running the commands from `packages/client` and doing:

`npx eslint src --fix`
`npx prettier src --write`

of course doing commands on `src` will be a huge change - i imagine we will slowly get our files formatted overtime as we work on them.

KsenijaZivkovic:
> @KsenijaZivkovic which command were you trying? the following should work after a `yarn install` (I only have `yarn` and `create-react-app` installed globally)
> 
> I was running the commands from `packages/client` and doing:
> 
> `npx eslint src --fix` `npx prettier src --write`
> 
> of course doing commands on `src` will be a huge change - i imagine we will slowly get our files formatted overtime as we work on them.

It's working for me, yesterday I forgot to navigate to the correct package, that's why it was failing.
I see that after the fix - all the quotes are changed to single quotes, if we want to keep the double quotes we need to disable the singleQuote

mannynotfound:
@KsenijaZivkovic I think we want single quotes, its the standard CAST uses, but we just have the rule set to "off" right now so it doesnt throw errors while developing: https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/.eslintrc.json#L31

KsenijaZivkovic:
> @KsenijaZivkovic I think we want single quotes, its the standard CAST uses, but we just have the rule set to "off" right now so it doesnt throw errors while developing: https://github.com/DapperCollectives/VESSEL/blob/main/packages/client/.eslintrc.json#L31

https://github.com/DapperCollectives/VESSEL/pull/126/files#diff-dbbe2fad4b33cee286a676401a44d78e2bfa368423a07d7142da3a313aa9417eR61 - i was referring to this line
I'm okay with single quotes, but I'm trying to recall that we had a team agreement on double quotes, was it after the JS guide meeting, can't remember correctly 🙈 
@amyliu6 @linxiao-li what do you say?

linxiao-li:
I made the decision to disable the single quote rule because double quotes are more prevalent in our codebase, and there's no significant advantage switching to use single quote only. I'd say let's unify and enforce the double quotes standard instead

mannynotfound:
@linxiao-li since there's no significant advantage to one or other, shouldn't we use what CAST has already implemented? Kind of odd to have that be the only differing rule in our codebases. Even though we have more double quotes at the moment, they are easily converted to single quotes with `--fix` or `formatOnSave`

KsenijaZivkovic:
> @KsenijaZivkovic re: pre-commit hook - I'm not sure if its been working since we didn't have a `.prettierrc` file before? the `pretty-quick` docs says it looks up the directory for a `.prettierrc`. In theory it should just work now - https://www.npmjs.com/package/pretty-quick

I'm not sure either if it was working before

KsenijaZivkovic (approved):

amyliu6:
> > @KsenijaZivkovic re: pre-commit hook - I'm not sure if its been working since we didn't have a `.prettierrc` file before? the `pretty-quick` docs says it looks up the directory for a `.prettierrc`. In theory it should just work now - https://www.npmjs.com/package/pretty-quick
> 
> I'm not sure either if it was working before

it was working before, we didn't define `.prettierrc` as the default setting matches with our format style, it's needed if we want to override the settings.

linxiao-li:
@mannynotfound I'm fine with that, just make sure we do the conversion in a single pr just for that purpose (possibly include in this pr?) just so the quote change doesn't get mixed with other feature/bug code changes

PR #128, opened 2022-10-06
VSL-121: Deposit Token Flow

KsenijaZivkovic:
## Description

Modal for depositing tokens was added
Works on clicking all deposit buttons in the app (on the home screen, and on the tokens page)
A button for deposit on Tokens page by selecting vault is only enabled if the user has that vault in their wallet.
Send button on Tokens page is disabled if the vault is empty

## Demo / Test Result

Screen Shot 2022-10-07 at 12 38 20 AM

Screen Shot 2022-10-07 at 12 38 46 AM

Screen Shot 2022-10-12 at 11 43 40 AM

In case of error, when user tries to send all the FLOW tokens to the treasury:
Screen Shot 2022-10-07 at 1 17 16 PM
Improvement created - https://linear.app/dappercollectives/issue/VSL-256

Tooltip:
Screen Shot 2022-10-07 at 1 18 09 PM
Improvement created - https://linear.app/dappercollectives/issue/VSL-254


linxiao-li:
1. Suggest renaming the directory and files to reflect the update 
`src/components/SendToken` (only send token flow)-> `src/components/TransferToken` (both send and deposit flow)
`src/components/SendToken/SendTokenContext` -> `src/components/SendToken/TransferTokenContext`

2. DepositToken.js and SendToken.js is the same except for the safedata and userAddress properties, and those two can be accessed anywhere in the child component using the web3 context, suggest combine them into one `TransferTokenContext`

3. Instead of doing a bunch of conditional rendering in the child components (i.e. ` if(transactionType === TRANSACTION_TYPE.SEND){...}`), can we try to abstract the interface for the child component and pass executables and properties? That can help reducing the complexity of the code

linxiao-li (commented):
A lot of changes are prettier quote changes, @mannynotfound can you create a pr that takes care of the quote changes all together? We can merge that one first. That will help making this pr cleaner

amyliu6:
@KsenijaZivkovic  re ```Send button on Tokens page is hidden if the vault is empty``` , wasn't aware that we need to hide 'send' button when vault is empty, could you help clarify if that's a new design change request from Alaa?

KsenijaZivkovic:
> @KsenijaZivkovic re `Send button on Tokens page is hidden if the vault is empty` , wasn't aware that we need to hide 'send' button when vault is empty, could you help clarify if that's a new design change request from Alaa?

@amyliu6 I have proposed this change, because I feel that it shouldn't be possible to send tokens from the empty vault. Even if the user opens the vault they wouldn't be able to change the amount and BE also prevents the users to propose 0 tokens to account. Let me know your thoughts, I can easily revert it!

amyliu6:
> > @KsenijaZivkovic re `Send button on Tokens page is hidden if the vault is empty` , wasn't aware that we need to hide 'send' button when vault is empty, could you help clarify if that's a new design change request from Alaa?
> 
> @amyliu6 I have proposed this change, because I feel that it shouldn't be possible to send tokens from the empty vault. Even if the user opens the vault they wouldn't be able to change the amount and BE also prevents the users to propose 0 tokens to account. Let me know your thoughts, I can easily revert it!

That true. As there were design exploration to notify user about empty vault in "Send flow" (for future release), hence want to double check here, we should only update design after getting their approval.

KsenijaZivkovic:
> A lot of changes are prettier quote changes, @mannynotfound can you create a pr that takes care of the quote changes all together? We can merge that one first. That will help making this pr cleaner

@linxiao-li PR created: https://github.com/DapperCollectives/VESSEL/pull/130

packages/client/src/components/SafeHome.js
```diff
@@ -1,20 +1,26 @@
...
   onApprove,
   onReject,
   onExecute,
 }) => (
   <>
-    <SafeOverview allBalance={allBalance} allNFTs={allNFTs} />
+    <SafeOverview
+      allBalance={allBalance}
+      allNFTs={allNFTs}
+      safeData={safeData}
```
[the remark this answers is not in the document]
KsenijaZivkovic: safeData can be accessed if we pass through the address, I will create a commit soon so you can take a look

packages/client/src/components/SafeNFTs.js
```diff
@@ -58,20 +58,23 @@ function SafeNFTs({ web3, address }) {
...
+            type="button"
             className="button is-primary flex-1"
             onClick={() =>
               modalContext.openModal(
                 <SendTokens
                   address={address}
                   initialState={{
                     assetType: ASSET_TYPES.NFT,
-                    selectedNFT: collection.key + "-" + token.tokenId,
+                    selectedNFT: `${collection.key}-${token.tokenId}`,
```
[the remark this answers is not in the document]
KsenijaZivkovic: Let me know if I changed it correctly!

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
```diff
@@ -0,0 +1,101 @@
...
+
+  useEffect(() => {
+    if (!userAddress) {
+      return;
+    }
+    getVaults();
+    // eslint-disable-next-line
+  }, [userAddress]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
KsenijaZivkovic: > > > @KsenijaZivkovic re `Send button on Tokens page is hidden if the vault is empty` , wasn't aware that we need to hide 'send' button when vault is empty, could you help clarify if that's a new design change request from Alaa?
> > 
> > 
> > @amyliu6 I have proposed this change, because I feel that it shouldn't be possible to send tokens from the empty vault. Even if the user opens the vault they wouldn't be able to change the amount and BE also prevents the users to propose 0 tokens to account. Let me know your thoughts, I can easily revert it!
> 
> That true. As there were design exploration to notify user about empty vault in "Send flow" (for future release), hence want to double check here, we should only update design after getting their approval.

Updated as per design, to disable buttons instead!

packages/client/src/components/SendTokens/steps/SendTokenForm.js
```diff
@@ -1,23 +1,28 @@
...
+import AddressDropdown from '../components/AddressDropdown';
+import AmountInput from '../components/AmountInput';
+import AssetSelector from '../components/AssetSelector';
+import { ASSET_TYPES } from 'constants/enums';
+import { SendTokensContext } from '../sendTokensContext';
 
 const SendTokenForm = () => {
   const [sendModalState] = useContext(SendTokensContext);
-  const { assetType } = sendModalState;
+  const { assetType, address } = sendModalState;
```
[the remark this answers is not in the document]
KsenijaZivkovic: In SendTokenForm I am fetching the balances now from the Web3 context
But in deposit form the balances of the user's vault should be fetched each time the deposit token is opened, to make sure the balances are correct (it can change every second, and it is not controlled by Vessel)

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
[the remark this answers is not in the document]
KsenijaZivkovic: > Can you help me understand the purpose of this effect? Looks like it will only be called on the deposit token modal open since vaults won't change during the depositing flow?

It was the only way to sync treasury vaults and user vaults -> to fetch the vaults from the treasury first and then to update the vaults of the user that can be used to deposit tokens.

packages/client/src/components/SendTokens/components/ButtonGroup.js
```diff
@@ -41,22 +47,38 @@ const ButtonGroup = () => {
...
-          tokenAmount,
-          coinType
-        );
-      } else {
-        await web3.proposeNFTTransfer(
-          formatAddress(address),
-          formatAddress(recipient),
-          `${collectionName}-${tokenId}`
+      try {
+        if (assetType === ASSET_TYPES.TOKEN) {
```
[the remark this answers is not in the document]
KsenijaZivkovic: I've extracted the function for transferring tokens out of button group

KsenijaZivkovic:
> > > @KsenijaZivkovic re `Send button on Tokens page is hidden if the vault is empty` , wasn't aware that we need to hide 'send' button when vault is empty, could you help clarify if that's a new design change request from Alaa?
> > 
> > 
> > @amyliu6 I have proposed this change, because I feel that it shouldn't be possible to send tokens from the empty vault. Even if the user opens the vault they wouldn't be able to change the amount and BE also prevents the users to propose 0 tokens to account. Let me know your thoughts, I can easily revert it!
> 
> That true. As there were design exploration to notify user about empty vault in "Send flow" (for future release), hence want to double check here, we should only update design after getting their approval.

Updated according to the designs, buttons are disabled instead of hidden!

packages/client/src/hooks/useAccount.js
```diff
@@ -37,35 +42,70 @@ export default function useAccount() {
...
+        }
       } catch (error) {
         console.log(`error getting balance for ${coin.coinType}`, error);
       }
     }
-    return result;
+    return result.filter((balance) => balance.balance);
+  };
+
+  const getUserVaults = async (address) => {
```
amyliu6, line 86: do we need `getUserVaults` func? can we use `getUserBalances` and check if `balance is > 0` to know whether vault exist?

packages/client/src/hooks/useAccount.js
```diff
@@ -1,23 +1,28 @@
...
 import { mutate, query, tx } from '@onflow/fcl';
-import { GetAccountBalanceByContractName, SEND_TOKENS_TO_TREASURY } from 'flow';
+import {
+  GET_ACCOUNT_BALANCE,
+  SEND_TOKENS_TO_TREASURY,
+  VAULT_EXIST_CHECK,
+} from 'flow';
 
-const COIN_TYPE_LIST = [COIN_TYPES.FLOW, COIN_TYPES.FUSD, COIN_TYPES.USDC];
+const COIN_TYPE_LIST = Object.keys(COIN_TYPE_TO_META);
```
amyliu6, line 10: if it's now entire list maybe this line can be removed and directly use it inside method.

packages/client/src/hooks/useAccount.js
KsenijaZivkovic: Vault can exist with 0 balance, and it's difficult to differentiate if the vault exists or not if both return 0 as a result - but in our use case it's okay to do it in a way you proposed!

packages/client/src/components/TransferToken/components/ButtonGroup.js
```diff
@@ -22,8 +24,12 @@ const ButtonGroup = () => {
     recipientValid,
     selectedNFT,
     coinType,
+    transactionType,
   } = sendModalState;
 
+  const safeAddress =
+    transactionType === TRANSACTION_TYPE.SEND ? address : recipient;
```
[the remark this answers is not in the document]
KsenijaZivkovic: Address in this case is the sender address, so in case of sending tokens - it's a treasury address, and in case of depositing -> it's the user's address. I will rename this to sender, so it will make more sense!

packages/client/src/flow/getAccountBalanceByContractName.tx.js
```diff
@@ -1,16 +1,21 @@
...
         let vaultCap = getAccount(address).getCapability<&${contractName}.Vault{FungibleToken.Balance}>(vaultPath)
-        if vaultCap.check() {
-            let vaultRef: &${contractName}.Vault{FungibleToken.Balance} = vaultCap.borrow()??panic("Could not borrow Balance reference to the Vault")
-            return vaultRef.balance
-        } else{
-            return 0.00
-        }
-    }`;
+        return vaultCap.check() 
+    }
```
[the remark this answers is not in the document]
KsenijaZivkovic: I can revert it to first version if we go with the vault > 0 for depositing!

packages/client/src/hooks/useAccount.js
```diff
@@ -37,35 +42,70 @@ export default function useAccount() {
...
+            address
+          );
+          coin['balance'] = coinBalance;
+        }
       } catch (error) {
         console.log(`error getting balance for ${coin.coinType}`, error);
       }
     }
-    return result;
+    return result.filter((balance) => balance.balance);
```
amyliu6, line 83: `result.filter((coin) => coin.balance)` for readability

packages/client/src/hooks/useAccount.js
KsenijaZivkovic: I would leave it like this if that is okay, because this variable is used in a couple of places so might be easier to read the code

packages/client/src/hooks/useAccount.js
KsenijaZivkovic: On page where we see treasury vaults -> we need the information if we should enable/disable the button for depositing tokens, there we don't need info about balance, but rather if the vault exists or not on the user side -> so that way we avoid more requests than needed (the number of requests doubles if we call it on that page).

amyliu6: make sense. thanks for explaining to me!

amyliu6:
@KsenijaZivkovic  has mobile web UI been implemented in this PR? if so could you please attach screenshot for that too?

KsenijaZivkovic:
> @KsenijaZivkovic has mobile web UI been implemented in this PR? if so could you please attach screenshot for that too?

No, currently both send/deposit modals aren't mobile responsive

packages/client/src/components/TransferToken/steps/index.js
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  return (
+    <div className="has-text-black">
+      <ModalHeader title={title} />
+      {currentStep === 1 ? (
+        <TransferTokenConfirmation />
+      ) : (
+        renderTransferTokenForm()
+      )}
```
linxiao-li, line 39:
```suggestion
{currentStep === 1 && <TransferTokenConfirmation/>}
{currentStep ===0 && isSendTransaction &&<SendTokenForm/>}
{currentStep === 0 && !isSendTransaction &&<DepositTokenForm/>}
```

packages/client/src/components/TransferToken/components/TransactionDetails.js
```diff
@@ -1,24 +1,54 @@
...
+      label = 'Sent';
+      break;
+    case TRANSACTION_TYPE.DEPOSIT:
+      senderName = getNameByAddress(contacts, sender);
+      recipientName = safeName;
+      label = 'Deposit';
+      break;
+    default:
+      break;
+  }
```
mannynotfound, line 48: since theres only 2 conditions this would probably be clearer as an if/else statement - unless there is other transaction types coming in which case switch is probably better

packages/client/src/components/TransferToken/steps/DepositTokenForm.js
```diff
@@ -0,0 +1,106 @@
...
+      console.log(`Failed to get coin balances, error: ${err}`);
+    }
+  };
+
+  useEffect(() => {
+    if (!userAddress) {
+      return;
+    }
+    getVaults();
+    // eslint-disable-next-line
```
mannynotfound, line 53:
```suggestion
   // eslint-disable-next-line react-hooks/exhaustive-deps
```

we should probably get out of the habit of abusing this eslint ignore though, apparently its not _good code_ https://dev.to/wkrueger/never-ignore-the-exhaustive-deps-rule-2ap8

packages/client/src/hooks/useAccount.js
```diff
@@ -1,71 +1,107 @@
...
+  const getUserVaults = async (address) => {
+    const result = COIN_TYPE_LIST.map((coinType) => ({
+      coinType,
+    }));
+    for await (const coin of result) {
+      const { coinType } = coin;
+      try {
+        const { contractName } = COIN_TYPE_TO_META[coinType];
+        const vaultPath = COIN_TYPE_TO_META[coinType].publicBalancePath;
+        coin['exists'] = await getVaultExists(contractName, vaultPath, address);
```
mannynotfound, line 91:
```suggestion
        coin.exists = await getVaultExists(contractName, vaultPath, address);
```

packages/client/src/modals/Transaction/BannerInfo.js
```diff
@@ -22,7 +22,7 @@ const BannerInfo = ({
   const { displayName, tokenType } = getTokenMeta(vaultId);
   const { contractName: NFTName, contractAddress: NFTAddress } =
     parseIdentifier(collectionId);
-  const { name: imageName, thumbnail: imageURI } = nftMeta || {};
+  const { name: imageName, imageURI } = nftMeta || {};
```
mannynotfound, line 25: this should conform to `MetadataViews.cdc` and use `thumbnail` like in this PR: https://github.com/DapperCollectives/VESSEL/pull/131

mannynotfound (commented):
looks pretty good to me! 

metadata image field needs to be changed + a few suggestions

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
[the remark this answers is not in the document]
linxiao-li: I guess I'm missing the point of having two separate useEffect hooks, why not combine with the hook above?

packages/client/src/components/TransferToken/components/TransactionDetails.js
```diff
@@ -1,24 +1,54 @@
...
-  const { safeOwners } = web3?.treasuries?.[address];
-  const recipientName =
-    safeOwners.find((owner) => owner.address === recipient)?.name ?? '';
   const clipboard = useClipboard();
+  const web3 = useContext(Web3Context);
+  const safeAddress =
+    transactionType === TRANSACTION_TYPE.SEND ? sender : recipient;
+
+  const { contacts } = useContacts(safeAddress);
+  const safeName = web3?.treasuries?.[safeAddress].name;
```
linxiao-li, line 30:
```suggestion
  const safeName = web3?.treasuries?.[safeAddress]?.name;
```

packages/client/src/components/SendTokens/steps/SendTokenForm.js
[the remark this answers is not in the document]
linxiao-li: What I mean is more about where to put the data after fetching (either from web3 context or from api call for user's vaults), right now we fetch data on `SendTokenForm` and `DepositTokenForm` then pass them down through the property chain via property `coinBalances`, what I'm suggesting is putting the coinBalances data into the context, and have cointypedropdown reading from the context. Let me know if it makes more sense.

packages/client/src/components/TransferToken/components/CoinTypeDropDown.js
```diff
@@ -1,17 +1,12 @@
...
-  const { balances } = web3;
-  const balanceMap = balances[address];
-  const coinTypes = Object.entries(balanceMap).map(([key, value]) => ({
+const CoinTypeDropDown = ({ coinType, setCoinType, balances, tooltipText }) => {
+  const coinTypes = Object.entries(balances).map(([key]) => ({
     itemValue: key,
     displayText: key,
     attr: {
-      balance: balanceMap[key] ? Number(balanceMap[key]).toFixed(2) : 0,
+      balance: balances[key] && Number(balances[key]).toFixed(2),
```
linxiao-li, line 9: What's the intention for the change here? Do we want to show an empty string or 0 when the balances[key] is undefined?

packages/client/src/components/SafeNFTs.js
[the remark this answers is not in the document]
linxiao-li: All good!

packages/client/src/components/SendTokens/steps/SendTokenForm.js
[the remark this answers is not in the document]
KsenijaZivkovic: Ohhh right, makes sense! I will put it inside the context and fetch it instead of passing through property. Thank you!

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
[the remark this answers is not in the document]
KsenijaZivkovic: I'll try again and change it if it works in one useEffect hook.

packages/client/src/components/TransferToken/components/TransactionDetails.js
KsenijaZivkovic: No, it's only two of them, I'll update it!

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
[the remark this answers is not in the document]
KsenijaZivkovic: It causes infinite loops when merged into one useEffect. The use case that this is solving is when user opens the page for the first time and local storage is empty, when vaults are loaded to update the user balance

packages/client/src/components/TransferToken/TransferTokensContext.js
```diff
@@ -0,0 +1,24 @@
...
+    selectedNFT: initialState?.selectedNFT ?? {},
+    transactionType: initialState?.transactionType ?? TRANSACTION_TYPE.SEND,
+  };
+  const [sendModalState, setSendModalState] = useState(INIT_STATE);
+  return (
+    <TransferTokensContext.Provider value={[sendModalState, setSendModalState]}>
+      {children}
+    </TransferTokensContext.Provider>
+  );
+};
```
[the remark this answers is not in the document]
KsenijaZivkovic: Makes sense to put it inside contexts, even though it's only used here

packages/client/src/modals/Transaction/BannerInfo.js
KsenijaZivkovic: It doesn't work if I change it to thumbnail, the banner doesn't show an image.
If we make it work for ExampleNFT, how will it behave on testnet for all the NFTs that are deployed there?
Examples can be found here: https://nft-catalog.vercel.app/catalog/testnet/Momentables

packages/client/src/components/TransferToken/components/CoinTypeDropDown.js
KsenijaZivkovic: I'm not sure, I'll revert it!

packages/client/src/modals/Transaction/BannerInfo.js
mannynotfound: @KsenijaZivkovic this should work for both I think:

`const imageURI = nftMeta?.imageURI ?? nftMeta?.thumbnail;`

mannynotfound (approved):

packages/client/src/components/SendTokens/steps/DepositTokenForm.js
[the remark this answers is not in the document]
linxiao-li: can we do something like this?
```
useEffect(()=>{
    if(isEmpty(vaults)){
         getTreasuryVaults();
    }
    updateUserBalance();
}, [userAddress])
```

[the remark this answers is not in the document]
KsenijaZivkovic: Updated it like this:
```
  useEffect(() => {
    if (!userAddress) {
      return;
    }
    if (isEmpty(vaults)) {
      getVaults();
    } else {
      updateUserBalance();
    }
  }, [userAddress, vaults]);
  ```

linxiao-li (approved):
LGTM

PR #131, opened 2022-10-13
replace zeedzino with our own exampleNFT contract

mannynotfound:
## Description

I had these changes locally for debugging but I think we should move to using our own example NFT contract that we own vs ZeedzINO. Additionally this fixes some calls in the hooks to properly get the image of the NFT based on metadata views.


packages/client/src/components/TestToolBox.js
```diff
@@ -18,7 +18,7 @@ const TestToolBox = ({ address }) => {
   };
 
   const onDepositNFT = async () => {
-    await sendNFTToTreasury(address, 0);
+    await sendNFTToTreasury(address, 3);
```
mannynotfound, line 21: so this magic number 3 is the token ID jackson sent me - we should probably make this dynamic or something otherwise you'll always have to edit this number depending on the token id you have

KsenijaZivkovic:
LGTM, but I would maybe leave ZeedzINO to exist along with ExampleNFT, in case if we delete it - all the transactions and tests on BE will need to be reverted to ExampleNFT because they will fail otherwise

linxiao-li:
I think we should keep zeedzIno as well, it's nice to have multiple collections for FE to test

packages/client/src/components/TestToolBox.js
jacksonConrad: just need to run the proper transaction.  as Ksenija noted, our `mint_nft` tx was updated from ExampleNFT -> ZeedzINO.  Here is a transaction that you can use to mint an example NFT.  Simply run this transaction from the service account to mint an NFT and deposit it to the `recipient` account.  For example:

```
flow transactions send ./transactions/mint_example_nft.cdc 0xf8d6e0586b0a20c7 "name" "description" "thumbnail"
```

mint_example_nft.cdc
```
import NonFungibleToken from "../contracts/core/NonFungibleToken.cdc"
import ExampleNFT from "../contracts/core/ExampleNFT.cdc"
import MetadataViews from "../contracts/core/MetadataViews.cdc"

/// This script uses the NFTMinter resource to mint a new NFT
/// It must be run with the account that has the minter resource
/// stored in /storage/NFTMinter
transaction(
    recipient: Address,
    name: String,
    description: String,
    thumbnail: String
) {

    /// local variable for storing the minter reference
    let minter: &ExampleNFT.NFTMinter

    /// Reference to the receiver's collection
    let recipientCollectionRef: &{NonFungibleToken.CollectionPublic}

    /// Previous NFT ID before the transaction executes
    let mintingIDBefore: UInt64

        self.mintingIDBefore = ExampleNFT.totalSupply

        // borrow a reference to the NFTMinter resource in storage
        self.minter = signer.borrow<&ExampleNFT.NFTMinter>(from: ExampleNFT.MinterStoragePath)
            ?? panic("Account does not store an object at the specified path")

        // Borrow the recipient's public NFT collection reference
        self.recipientCollectionRef = getAccount(recipient)
            .getCapability(ExampleNFT.CollectionPublicPath)
            .borrow<&{NonFungibleToken.CollectionPublic}>()
            ?? panic("Could not get receiver reference to the NFT Collection")
    }

    execute {

        // Mint the NFT and deposit it to the recipient's collection
        self.minter.mintNFT(
            recipient: self.recipientCollectionRef,
            name: name,
            description: description,
            thumbnail: thumbnail,
            royalties: [] 
        )
    }

    post {
        self.recipientCollectionRef.getIDs().contains(self.mintingIDBefore): "The next NFT ID should have been minted and delivered"
        ExampleNFT.totalSupply == self.mintingIDBefore + 1: "The total supply should have been increased by 1"
    }
}
```

You can change up the first argument to deposit the NFT into another account's storage, they just need to have an `ExampleNFT.Collection` created in their storage already.  Here is the tx to do that:

```
import NonFungibleToken from "../contracts/core/NonFungibleToken.cdc"
import ExampleNFT from "../contracts/core/ExampleNFT.cdc"

// This transaction is what an account would run
// to set itself up to receive NFTs
transaction {
        // Return early if the account already has a collection
        if signer.borrow<&ExampleNFT.Collection>(from: ExampleNFT.CollectionStoragePath) != nil {
            panic("found collection!")
        }

        // Create a new empty collection
        let collection <- ExampleNFT.createEmptyCollection()

        // save it to the account
        signer.save(<-collection, to: ExampleNFT.CollectionStoragePath)

        // create a public capability for the collection
        signer.link<&{NonFungibleToken.CollectionPublic, ExampleNFT.ExampleNFTCollectionPublic}>(
            ExampleNFT.CollectionPublicPath,
            target: ExampleNFT.CollectionStoragePath
        )
    }
}
```

mannynotfound: @jacksonConrad @amyliu6 added these transactions, kept the zeedz transactions, and renamed each `mint_example_NFT`/`mint_zeedz_NFT` and `create_example_NFT_collection`/`create_zeedz_NFT_collection` respectively and updated our tests to use exampleNFT

jacksonConrad (approved):

amyliu6 (approved):

PR #132, opened 2022-10-14
VSL-224: Modal improvements

KsenijaZivkovic:
## Description

Reused BannerInfo component and deleted unused components
This resulted in slightly smaller font for add/remove owner and send tokens modal (it was too big before, and now it's aligned with transactions modal"

## Demo / Test Result

Screen Shot 2022-10-14 at 3 22 40 PM

### DIFF
--- a/packages/client/src/components/SafeNFTs.js
+++ b/packages/client/src/components/SafeNFTs.js
@@ -37,6 +37,9 @@ function SafeNFTs({ web3, address }) {
     .filter((x) => x);
 
   const showNFTModal = (collection, token) => {
+    const { collectionName, key: collectionId } = collection;
+    const { tokenId } = token;
+
     modalContext.openModal(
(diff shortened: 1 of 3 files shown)


packages/client/src/components/SafeNFTs.js
```diff
@@ -73,7 +76,7 @@ function SafeNFTs({ web3, address }) {
                   sender={address}
                   initialState={{
                     assetType: ASSET_TYPES.NFT,
-                    selectedNFT: token,
+                    selectedNFT: { ...token, collectionName: collectionId },
```
KsenijaZivkovic, line 79: Token doesn't have information about a collection, so I added it here to make sending tokens from NFTs page work. When user opens the modal, the NFT will be selected, and there will be no errors when this NFT is sent to the account.

PR #135, opened 2022-10-31
Typescript

linxiao-li:
## Description
- Introduces Typescript into VESSEL project, added dependencies and Eslint configurations to accommodate typescript and javascript
- Rewrites the reusable components with typescript
- Adjust the param passing for the dropdown renderOption function to align with typescript best practices

## Test on your device
1. run `rm -rf node_modules` in both root dir and `packages/client`
2. delete all lock files
3. run `npm cache clean -f`
4. run yarn in root dir


packages/client/src/library/components/InputAddress.tsx
```diff
@@ -4,14 +4,31 @@ import { useAddressValidation } from 'hooks';
...
 import Svg from 'library/Svg';
 
-const InputAddress = ({ web3, value, isValid, onChange }) => {
+export type AddressValidationResult = {
+  value: string;
+  isValid: boolean;
+};
+
+export interface InputAddressProps {
+  web3: any;
```
linxiao-li, line 13: Giving the web3 an any type since we are just starting the migration, there will be followup work to define the type for web3

